Issue #31, opened 2019-11-12
pybind11 folder rework


simcof:

chinmayshah99:
I'll take this up

Issue #7, opened 2019-11-10
Convert SWIG to pybind11


simcof:
Update initial codebase and set up scripts to pybind11

chinmayshah99:
@simcof , we can close this issue.

simcof:
done

PR #33, opened 2019-12-04
DP library ported to Cmake

chinmayshah99:
Ported the DP library to Cmake. Major steps added
- added boringssl and Go installation (required by DP library)
- Shited submodule HEAD of Google DP library to [here- fork of repo](https://github.com/chinmayshah99/differential-privacy) as it maintains the CMake build system. (Else recursive submodule init was breaks down). 

#31 #7

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,14 @@ build/
 __pycache__/
 *.py
 !setup.py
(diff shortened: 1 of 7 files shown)


simcof:
CMake conversion

PR #39, opened 2020-02-18
remove cmakelits and install go

simcof:
remove cmakelits and install go


simcof:
remove cmakelits and install go, this is a leftover task from merge

Issue #57, opened 2020-02-23
Add support for python 2 and 3 in build config


simcof:
Build config currently defaults to building for python 2 (usually). Need to support multiple configuration

TTitcombe:
Out of interest, why support 2 and 3, given that Python 2 is end of life?

chinmayshah99:
My bad. 
So the idea was in the system with just python3 installed, we can call python by simple `python`, but in the systems with both python 2 and 3, we needed to call python as `python3`. 
Apologies for not making it clear.

TTitcombe:
That makes sense, thanks for clearing that up

PR #63, opened 2020-02-26
Fixed bug in environment setup-process

chinmayshah99:
While fixing #57 , it broke support in system with only python3 installed.


chinmayshah99:
oops, made to wrong base :(

Issue #41, opened 2020-02-18
Establish Python package framework


simcof:
- establish .cpp bindings (src/bindings) structure
- establish python extension API

chinmayshah99:
For `src/bindings`, we should follow the same structure as Google DP library.

chinmayshah99:
While we have ported the CPP functions, we need to decide on how the python library will be exactly used. i.e., python API structure and how will the functions like `serialize()` and `partial output()` interact.

alejandrosame:
Should this issue be closed or updated with new requirements?

chinmayshah99:
While #291 talks about organizing the source code, we also need to identify how will we organize the python modules, as we add more modules. Are we going to follow `algorithms.__noise_mechanism__.function` or should we think of a different approach? 

But for all practical purposes at least for now, this issue does not serve utility.

PR #68, opened 2020-03-03
Adding pydp.so to python wheel

vishalsubbiah:
Adding the .so file similar to how tensorflow adds it.
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/tools/pip_package/setup.py#L280

issue: https://github.com/OpenMined/PyDP/issues/41
cc @simcof

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,6 +38,9 @@ setup(
     include_package_data=True,
     keywords='pydp',
     name='pydp',
+    package_data={
+        'pydp': ['pydp.so'],
+    },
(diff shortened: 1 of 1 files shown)


chinmayshah99:
Hi @vishalsubbiah , are you able to install the package and run `import pydp` in python3?

simcof:
This adds pydp.so to the wheel, but does not allow us to access the pydp package once we install the wheel with pip. Let's get this working before we merge. Also we need to assign a related issue

vishalsubbiah:
moved changes to #71

PR #73, opened 2020-03-06
Add continuous integration via Github Actions

### LINKED ISSUE #56: Maintain Python3 support in all files
In `Makefile `and `tox.ini`, the command invoked is `python file.py`.

The problem is in `conda `env, it defaults to python3. But in systems with both python 2 and 3 installed, it calls python2. 

#### What can be done:
Make sure the correct python version is invoked when you call `python` and make changes accordingly in these two files.

Benardi:
## Add continuous integration via Github Actions

Add Github Actions workflow to allow Continuous Integration. Relay pip install to workflow.

Fixes #56 

## Type of change

## How Has This Been Tested?

- [x] py35, py36, py37 testenvironments
- [x] [Checks](https://github.com/Benardi/PyDP/runs/491223097) running on fork 

## Checklist:

- [ ] New Unit tests added
- [x] Unit tests pass locally with my changes

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,46 @@
+name: Tests
+
+on:
+  push:
+    branches:
+    - master, dev
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 4
+      matrix:
+        python-version: [3.5, 3.6, 3.7]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install python dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r ./requirements_dev.txt
+    - name: Set up submodules
+      run: |
+        bash ./ext_source_setup
+    - name: Install pre-requisites
+      run: |
+        sudo apt-get update
+        bash ./prereqs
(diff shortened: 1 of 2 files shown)


Benardi:
> Can we add support for tox tests?

It was my first thought actually. Haven't been able to get it to work with Github Actions, despite following the standard procedure. It seems to intermittently fail.

- [Checks on PR](https://github.com/Benardi/PyDP/pull/2/checks?check_run_id=491304862) for Branch using tox

- [workflow script](https://github.com/Benardi/PyDP/blob/d05c56bb6ad91ba917d9ccfdb03f23618ff346bc/.github/workflows/tests.yml) for tox version
 -  [Github help page recommended procedure](https://help.github.com/en/actions/language-and-framework-guides/using-python-with-github-actions#running-tests-with-tox)

chinmayshah99:
hey @Benardi ,
So we are using pytest as of now. If you see #74, tox is calling pytest. 
So if we can find a way around?

Benardi:
@chinmayshah99  You could call pytest directly in the workflow. Let me raise a PR to your fork's branch

Benardi:
@chinmayshah99  In regards to pytest, you could use pytest directly in the workflow file. There seems to be conflicts between your branch and dev regarding BINDINGS, I believe you would be better than me for dealing with this.

chinmayshah99:
@Benardi I have resolved that conflict. You're good to go!

chinmayshah99:
@Benardi , I think we no longer need to add anything. 

with reference to commit https://github.com/OpenMined/PyDP/pull/74/commits/89b55f14732f2d2a1f24d1de76dfdf34ec16e00d, all our tests are running good.

But it would be great if we can streamline our workflow, remove any redundancy(if any). 

Cheers!

PR #82, opened 2020-03-18
reduce the code

NileshPant1999:
## Description
 Reduce some line of code 

## Type of change

- [] New feature (non-breaking change which adds functionality)

### DIFF
--- a/tests/base/test_status.py
+++ b/tests/base/test_status.py
@@ -29,8 +29,8 @@ class TestStatus():
             "StatusCode.kUnauthenticated", 
         ]
         actual_codes = []
-        x = range(17)
-        for n in x:
+        
+        for n in range(17):
             actual_codes.append(str(pd.Status.StatusCode(n)))
         print(actual_codes)
         assert status_code_available == actual_codes
--- a/tests/base/test_status.py
+++ b/tests/base/test_status.py
@@ -30,8 +30,8 @@ class TestStatus():
         ]
         actual_codes = []
         
-        for n in range(17):
-            actual_codes.append(str(pd.Status.StatusCode(n)))
+        
+        [actual_codes.append(str(pd.Status.StatusCode(n))) for n in range(17)]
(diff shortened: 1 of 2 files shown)


tests/base/test_status.py
```diff
@@ -29,8 +29,8 @@ def test_status_codes(self):
             "StatusCode.kUnauthenticated", 
         ]
         actual_codes = []
-        x = range(17)
-        for n in x:
+        
+        for n in range(17):
```
jeamick: We can use List Comprehension if we really care about reducing the code size.
```suggestion
         actual_codes = [str(pd.Status.StatusCode(n)) for n in range(17)]
```
I guess since it is a test example what we want is more readability.

tests/base/test_status.py
```diff
@@ -29,8 +29,8 @@ def test_status_codes(self):
...
             actual_codes.append(str(pd.Status.StatusCode(n)))
```
jeamick:
```suggestion
           
```
Then we don't need this

jeamick:
```suggestion
           
```
So we don't need that

tests/base/test_status.py
```diff
@@ -30,8 +30,8 @@ def test_status_codes(self):
         ]
         actual_codes = []
```
jeamick:
```suggestion
        actual_codes = [actual_codes.append(str(pd.Status.StatusCode(n))) for n in range(17)]
```

tests/base/test_status.py
```diff
@@ -30,8 +30,8 @@ def test_status_codes(self):
...
         
-        for n in range(17):
-            actual_codes.append(str(pd.Status.StatusCode(n)))
+        
+        [actual_codes.append(str(pd.Status.StatusCode(n))) for n in range(17)]
```
jeamick:
```suggestion
        
```

PR #89, opened 2020-03-31
Add Dockerfile

### LINKED ISSUE #61: Add Dockerfile
Since the environment setup is very complex, using Docker would be the easiest way to setup. 
This will boost cross-platform development too, especially among our friends using Windows!

replomancer:
## Description

This Dockerfile defines an image with PyDP library. The building process follows the steps from contributing.md

Fixes #61 

## How Has This Been Tested?

When I run `make test-all` inside the container, the tests are working and I get the same errors and failures as on the host.:
```
E   AttributeError: module 'pydp' has no attribute 'status_code_to_string'
```
plus complaints about Python 3.5 and 3.7.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,34 @@
+# Pull base image.
+FROM python:3.6-slim-buster
+
+# Set environment variables.
+ENV HOME /root
+ENV PATH "/root/bin:${PATH}"
+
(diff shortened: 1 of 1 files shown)


chinmayshah99:
The error in `make test-all` has been fixed in #91

PR #118, opened 2020-04-29
File re-organization

chinmayshah99:
Major file organization and renaming variables to a standard. 

## How Has This Been Tested?

Yes, Bazel build passes

### DIFF
--- a/src/bindings/PyDP/algorithms/distributions.cpp
+++ b/src/bindings/PyDP/algorithms/distributions.cpp
@@ -6,31 +6,31 @@
 
 using namespace std;
 namespace py = pybind11;
-namespace dpi = differential_privacy::internal;
+namespace dpinternal = differential_privacy::internal;
 
 void declareLaplaceDistribution(py::module& m) {
-  py::class_<dpi::LaplaceDistribution>(m, "LaplaceDistribution")
+  py::class_<dpinternal::LaplaceDistribution>(m, "LaplaceDistribution")
       .def(py::init<double>())
-      .def("get_uniform_double", &dpi::LaplaceDistribution::GetUniformDouble)
+      .def("get_uniform_double", &dpinternal::LaplaceDistribution::GetUniformDouble)
       .def("sample",
-           (double (dpi::LaplaceDistribution::*)()) & dpi::LaplaceDistribution::Sample)
-      .def("sample", (double (dpi::LaplaceDistribution::*)(double)) &
-                         dpi::LaplaceDistribution::Sample)
-      .def("get_diversity", &dpi::LaplaceDistribution::GetDiversity)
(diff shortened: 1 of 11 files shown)


src/bindings/PyDP/algorithms/distributions.cpp
```diff
@@ -6,31 +6,31 @@
 
 using namespace std;
 namespace py = pybind11;
-namespace dpi = differential_privacy::internal;
+namespace dpinternal = differential_privacy::internal;
```
alejandrosame: I'm not against this, but if we go the route of only shortening differential_privacy namespace, I would completely leave behind renaming subnamespaces and just use dp::X where necessary.

alejandrosame (dismissed):
LGTM

alejandrosame (approved):

src/bindings/PyDP/algorithms/distributions.cpp
chinmayshah99: IMO
If we just stick to one namespace in a given file, we can shorten the namespace. 
But in case multiple namespaces are in the files, maybe stick to writing dp::X?

chinmayshah99: IMO
If we just stick to one namespace in a given file, we can shorten the namespace. 
But in case multiple namespaces are in the files, maybe stick to writing dp::X?

src/bindings/PyDP/bindings.cpp
```diff
@@ -10,24 +10,30 @@ void init_base_logging(py::module &);
 void init_base_percentile(py::module &);
 
 // algorithms
-void init_algorithms(py::module &);
+void init_algorithms_algorithms(py::module &);
```
alejandrosame: According to the changes, it should be `init_algorithms_bounded_functions`. This should fix the current test error.

alejandrosame (changes requested):
I didn't notice the error in `bindings.cpp` while reading the changes. I think that change should be enough to pass the tests again.

src/bindings/PyDP/bindings.cpp
chinmayshah99: Yeah, thanks! Making those changes.

src/bindings/PyDP/algorithms/distributions.cpp
chinmayshah99: hey @alejandrosame taken up your idea and made the changes in the latest commit.

alejandrosame (approved):
LGTM

src/bindings/PyDP/algorithms/distributions.cpp
alejandrosame: Ok! In any case we can always revise it later, if we find out that it interferes with readability for more complex bindings.

PR #123, opened 2020-05-03
Algorithm rand bindings

leclair-7:
## Description

Binds algorithms/rand to pydp. 

- [ x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?
I ran build PyDP.sh, ran python on my terminal:
import pydp as dp
dp.Geometric()
dp.UniformDouble()

Numbers matched what rand.h indicates they should be.


tests/algorithms/test_rand.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+import pydp as dp
+
+
+def test_rand_UniformDouble():
+    ud = dp.UniformDouble()
+    assert isinstance(ud, float)
```
chinmayshah99, line 7: https://github.com/google/differential-privacy/blob/master/differential_privacy/algorithms/rand_test.cc#L52
Can we add all of these tests?

chinmayshah99 (changes requested):

tests/algorithms/test_rand.py
leclair-7: Yeah good idea, I'm putting an isinstance assertion for every instance and function we currently have binded. I figure it's a decent sanity check on return policies.

chinmayshah99: While with these tests, we get to know that at least it is returning something sensible, we want to make sure that it doesn't return a junk value, hence tests for actual values.

leclair-7:
Ok, I've added instance checks to everything currently binded except utils. For some reason to import it, it has to be "import pydp.pydp.util" which I don't think is intended.

alejandrosame (dismissed):
I approve the PR, but I want to raise awareness that it is bad practice to make PRs (and generally group commits) that span files unrelated to the feature that the PR targets.

In this case, the addition of `isinstance` tests for the other classes should be made in a separate PR to make the scope clearer. It's not critical at all right now, but at some point I expect that we will need to reinforce this kind of workflows.

tests/algorithms/test_algorithms.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+import pydp as dp
+
+
+def test_algorithms_BoundedMean():
```
chinmayshah99: This function needs to be shifted to `test_bounded_mean.py`

chinmayshah99 (changes requested):

tests/algorithms/test_algorithms.py
chinmayshah99: hey @leclair-7, I meant moving it to the different file called `test_bounded_mean.py`

leclair-7: Oh gotcha

leclair-7: All good now

chinmayshah99 (approved):
LGTM!

Issue #59, opened 2020-02-26
Carrots demo


simcof:
Bind enough of the library to execute the carrots demo found in the examples folder of googles DP library

jeamick:
@jeamick

chinmayshah99:
Partial functionality added in https://github.com/OpenMined/PyDP/commit/36bd27b692c53173e1a4632948854bd2b95bf66c.

chinmayshah99:
Added as part of #117

PR #128, opened 2020-05-06
carrots demo

codeboy5:
## Description

This PR partially adds functions for the completing the carrots demo.

Partial issue #59  

## Type of change

## How Has This Been Tested?

Did you run `make test-all`? Yes I ran `make test`

## Checklist:

- [ ] New Unit tests added
- [ ] Unit tests pass locally with my changes

### DIFF
--- a/examples/carrots.py
+++ b/examples/carrots.py
@@ -55,6 +55,7 @@ class CarrotReporter:
         return x.result(list(self._df[self._df.carrots_eaten > limit]["carrots_eaten"]))
 
     def private_max(self, privacy_budget: float) -> dp.StatusOrO:
+        # 0 and 150 are the upper and lower limits for the search bound.
         x = dp.Max(privacy_budget, 0, 150)
         return x.result(list(self._df["carrots_eaten"]), privacy_budget)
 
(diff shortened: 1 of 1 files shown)


examples/carrots.py
```diff
@@ -51,16 +51,12 @@ def private_mean(self, privacy_budget: float) -> dp.StatusOrO:
...
-        pass
-
-    def private_count_above(self, privacy_budget: float, limit: int) -> dp.StatusOrO:
-        pass
+        x = dp.CountInt(privacy_budget)
+        return x.result(list(self._df[self._df.carrots_eaten > limit]["carrots_eaten"]))
 
     def private_max(self, privacy_budget: float) -> dp.StatusOrO:
-        pass
+        x = dp.Max(privacy_budget, 0, 150)
```
chinmayshah99, line 63: Please add a comment on what 0 and 150 represent

codeboy5: They are the search bounds, default values are numeric integers limits which gives a very inconsistent value.
Should I add a comment saying these are the search bounds ?

chinmayshah99: Yes, please do that. 
Also, can you please add it in other part of code if it's not present?

codeboy5: Okay, will do that

codeboy5: I've added the comment, the private sum requires similar bounds but we haven't ported that yet.

chinmayshah99 (approved):

Issue #119, opened 2020-04-29
Wheel tag doesn't match the actual package


youben11:
### Expected Behavior

The distributed wheel package is tagged with 'py2.py3-none-any' which means it should work on all platform with both python 2 and 3, however, the wheel is including the pydp.so file which is compiled for a specific platform and python3.6 exactly.

### Current Behavior

Can't run on a different platform or python other than 3.6

### Steps to Reproduce Behavior

Install pydp for other version than python3.6, or install it in a different platform than Linux (which the pydp.so file is compiled for)

chinmayshah99:
Similar to #113

chinmayshah99:
Fixed in #133

Issue #111, opened 2020-04-25
Workflow for publishing directly on PYPI


chinmayshah99:
### Expected Behavior

Create a Github workflow, which when a release is created, automatically builds the package and uploads it on pypi.

We will need to create three different workflows for Linux, OSX and windows.

### Current Behavior

What is the current behavior?
As of now, we have to manually upload the package when we are about to release them using twine on the local system.

chinmayshah99:
- [x] Linux
- [x] OSX
- [x] Windows

PR #136, opened 2020-05-11
Added pypi workflow 

### LINKED ISSUE #113: PyDP does not work on python 3,5, 3.7
### Expected Behavior

It should work on all python3 interpreters. 

### Current Behavior

When we import pydp on interpreters other than py3.6, It says module was compiled for python3.6; interpreter incompatible.

### Logs

image

### System information

* PyDP version: 0.1.0dev
* Python version: 3.5, 3.7

chinmayshah99:
## Description

Finished pypi auto-release via github actions

Fixes #113   

Partial issue #111 , #119  

## Type of change


.github/workflows/pypipublish_linux.yml
```diff
@@ -0,0 +1,67 @@
...
+              echo "Installing g++"
+              sudo apt-get install g++
+          fi
+          echo "Installing Bazel dependencies"
+          sudo apt-get install pkg-config zip zlib1g-dev unzip
+          echo "Donwloading Bazel 2.1.0"
+          wget https://github.com/bazelbuild/bazel/releases/download/2.1.0/bazel-2.1.0-installer-linux-x86_64.sh
+
+          chmod +x bazel-2.1.0-installer-linux-x86_64.sh
+          ./bazel-2.1.0-installer-linux-x86_64.sh --user
```
youben11, line 43: Do you require a specific version of bazel? [Bazel 3.0 is installed on Ubuntu 18.06](https://github.com/actions/virtual-environments/blob/master/images/linux/Ubuntu1804-README.md)

.github/workflows/pypipublish_linux.yml
```diff
@@ -0,0 +1,67 @@
...
+          python -m pip install --upgrade pip
+          pip install -r ./requirements_dev.txt
+
+      - name: Build wheel
+        run: |
+          python setup.py bdist_wheel
+
+      - name: Renaming wheel
+        run: |
+          find . -name '*linux*.whl' -type f -exec bash -c 'mv "$1" "${1/linux/manylinux1}"' -- {} \;
```
youben11, line 60: I don't agree that renaming is the way to go here, you should be using the auditwheel package for compliance, it might break on some systems.

.github/workflows/pypipublish_linux.yml
chinmayshah99: We have tested with this bazel version(2.1). Any upgrade in bazel versions needs to be tested

.github/workflows/pypipublish_linux.yml
chinmayshah99: The package is manylinux1 compliant. So adding it here. 
But in the meantime, we are in the process of looking for a permanent solution.

chinmayshah99:
Squashing this as it's 48 commits

.github/workflows/pypipublish_linux.yml
jeamick: Hi there ,

@chinmayshah99 let’s open an issue as it is a good practise to normally use auditwheel to publish on PyPi.

It will enable to keep track or the changes.

chinmayshah99: Yeah, let's do this. Are you creating an issue?

Issue #19, opened 2019-11-10
algorithms\bounded-standard-deviation


simcof:

replomancer:
I'm working on it.

PR #148, opened 2020-05-19
Add bindings for BoundedStandardDeviation

replomancer:
## Description

Adds bindings for BoundedStandardDeviation (through the C API).

Partial(?) solution of issue #19 

## Type of change

## How Has This Been Tested?

Did you run `make test-all`?
Some tests are still failing. New test is not very useful, but for more strict tests we need to start setting noise to zero.

## Checklist:

- [x] New Unit tests added
- [ ] Unit tests pass locally with my changes

### DIFF
--- a/tests/algorithms/test_bounded_standard_deviation.py
+++ b/tests/algorithms/test_bounded_standard_deviation.py
@@ -3,8 +3,10 @@ import pydp as dp
 
 class TestBoundedStandardDeviation:
     def test_c_api(self):
-        a = [1, 5, 7, 9, 13]
-
-        bsd = dp.BoundedStandardDeviation(1.0, 0, 15)
-        result = bsd.result(a)
+        example_data = [1, 5, 7, 9, 13]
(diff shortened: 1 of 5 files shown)


tests/algorithms/test_bounded_standard_deviation.py
```diff
@@ -0,0 +1,10 @@
+import pydp as dp
+
+
+class TestBoundedStandardDeviation:
+    def test_c_api(self):
```
chinmayshah99, line 5: Can we add a test on whether SD obtained is within a particular limit/ bound?

chinmayshah99 (dismissed):

tests/algorithms/test_bounded_standard_deviation.py
replomancer: Sure! I missed that. I also improved the test readability.

chinmayshah99 (approved):

PR #151, opened 2020-05-20
Include "C API" files in clang-format check

replomancer:
## Description

Apparently files in `src/bindings/c` are excluded from the style check and do not follow the formatting standard. Some other files avoided the check probably because of missing `-o` in the `find` command (details are in this PR style check results).

~This PR only fixes the style check (no files reformatted yet).~


chinmayshah99:
@replomancer Can we format these C API files using clang-format and add them to this PR?

chinmayshah99 (approved):

PR #155, opened 2020-05-25
Updated contributing.md and tested prereqs_mac.sh

madhavajay:
- Fixed pydp.so cleanup before new build

### DIFF
--- a/prereqs_mac.sh
+++ b/prereqs_mac.sh
@@ -3,17 +3,16 @@
 which -s brew
 if [[ $? != 0 ]] ; then
     # Install Homebrew
-    echo "Donwloading and installing homebrew"
+    echo "Downloading and installing homebrew"
     /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
 else
     brew update
 fi
 
-# TODO Python checking and installation
-# echo "Donwloading and installing Python3 homebrew"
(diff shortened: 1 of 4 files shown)


prereqs_mac.sh
```diff
@@ -3,17 +3,16 @@
...
     /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
 else
     brew update
 fi
 
-# TODO Python checking and installation
-# echo "Donwloading and installing Python3 homebrew"
-# brew install python
+echo "Downloading and installing Python3 using homebrew"
+brew install python3
```
chinmayshah99, line 13: Commented this out, because there can be conflicts. So we need to check if python is already installed, then skip this part.

prereqs_mac.sh
```diff
@@ -3,17 +3,16 @@
 which -s brew
 if [[ $? != 0 ]] ; then
     # Install Homebrew
-    echo "Donwloading and installing homebrew"
+    echo "Downloading and installing homebrew"
```
iamtrask, line 6: Nice touch! Really going the extra mile here!

iamtrask (approved):

prereqs_mac.sh
chinmayshah99: haha

PR #176, opened 2020-06-02
Docker testing publish

### LINKED ISSUE #163: Github actions for Dockerfile
#### Publish to Docker Hub
It would be great if we can push the dockerfile to the docker hub, so users can directly try it from docker hub.

#### Github Action Tests
At the same time, while Github actions do give us in an isolated environment to test our build, we want to check if our build actually doesn't break dockerfile & if possible run the container too to check if it actually works.

Action points:

- [ ] Push to Docker Hub
- [ ] Github Action to test it docker file builds

BenjaminDev:
# Pull Request

## Description
Publish to Docker Hub so it's easy for people to try out the library.

Changes:
- added a new workflow that builds the image, runs the test suit and (in theory - credentials will be needed) pushes the image to docker hub.
    
Notes: Dev dependencies are not baked into the image but only used to run tests. 

Fixes OpenMined#163
## Affected Dependencies
None

## Type of Change

- [ ] Documentation (non-breaking change which adds documentation)
- [x]  Improvement (non-breaking change that improves the performance or reliability of existing functionality)
- [ ] New feature (non-breaking change which adds functionality)

## How has this been tested?
- Github action is successful. 
- To reporduce: fork and push a change to `dev`. Note: make sure your github repo settings enables actions to run. Seems my defaults for forks in not to run the actions.  

## Checklist
- [ ] I did follow the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md)
- [x] I did follow the [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [ ] I have added tests for my changes

## Additional Context
Questions to conclude this PR
- When (which branches) should trigger this action? 
- What tagging convention should we use? (So far just latest and git sha) 
- What image naming should we use?
- We'll need to add a github secret with docker hub creds allow pushing once the rest has  been done.
- Should the tests we run on these images be against all supported python versions? I am unsure how the current tox setup is working but I'm sure we could run that. Depending on the intended use of these images I guess. 

I'm super happy to complete these but someone who knows those answers can pop them in and take over if they would like.


codeboy5:
I think we can have the publish workflow to run only a specific branch or tag, and the build needs to run on every PR to check the Docker-file

.github/workflows/update-docker-image.yml
```diff
@@ -0,0 +1,50 @@
...
+        run: docker build -t ${{steps.build_name.outputs.image_name}} --build-arg python_version=${{matrix.python-version}} -f Dockerfile .
+
+      - name: Tag images
+        run: |
+          docker tag ${{steps.build_name.outputs.image_name}} ${{steps.build_name.outputs.image_name}}:latest
+          docker tag ${{steps.build_name.outputs.image_name}} ${{steps.build_name.outputs.image_name}}:$GITHUB_SHA
+      - name: Run tests
+        run:
+          | # Not sure if this is what is desired. If a test fails the whole build fails - can mark flaky tests if needed?
+          docker run ${{steps.build_name.outputs.image_name}}:latest /bin/bash -c  "pipenv install --dev && pipenv run pytest ./tests"
```
chinmayshah99: This can be a separate workflow

.github/workflows/update-docker-image.yml
```diff
@@ -0,0 +1,50 @@
...
+    # tags:
+    #   - [v*]
+name: Update and Test Docker Image
+jobs:
+  publish:
+    name: Build and publish
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8]
```
chinmayshah99: Can we add python 3.5 too?

Dockerfile
```diff
@@ -1,5 +1,6 @@
 # Pull base image.
-FROM python:3.7-slim-buster
+ARG python_version
+FROM python:${python_version}-slim-buster
 
 # Set environment variables.
```
chinmayshah99: Can you resolve the merge conflict?
Our latest dockerfile deals with this

Dockerfile
```diff
@@ -50,7 +51,7 @@ RUN \
     cp -f ./bazel-bin/src/bindings/pydp.so ./pydp && \
     rm -rf dist/ && \
     pipenv run python3 setup.py bdist_wheel && \
-    pip3 install dist/*.whl
+    pipenv install dist/*.whl
```
chinmayshah99: If we do this. then when we open the bash, would the pipenv env open or the default python env?
The question is. if I type 
```
python3
>> import pydp as dp
```
Would this work?

chinmayshah99 (changes requested):

.github/workflows/update-docker-image.yml
zizu-research: Sure. I'd forgotten to check that.

BenjaminDev: Sure. I forgot to check that. Thanks

.github/workflows/update-docker-image.yml
BenjaminDev: I think it would be nicer to run the tests against the docker images before pushing them out the door. 
It won't add much to the total time taken to run the action. And the final shipped artifact will be identical.

Dockerfile
BenjaminDev: I believe we should install the final package into a virtual environment. If you don't want the user to have to run `pipenv shell` and to simply run `python` then I think this would work.
```
ENV PIPENV_VENV_IN_PROJECT=true 
# This `activates` the virtual env
ENV VIRTUAL_ENV=$PROJECT_DIR/.venv
ENV PATH="$VIRTUAL_ENV/bin:$PATH"
```
(edit: tested and it works)
However, to just install it outside a virtual env is not ideal and could have unforeseen issues. We could always add a line to the .bashrc file to put the user into the relevant environment?
Perhaps I need to full understand the use of these images and I can do my best to make them work as expected.

My current understanding of what we want out of this is: 
-  Publish docker images to Docker Hub so it's easy for people to try out the library.

BenjaminDev:
Remaining items:

1. Are the image names and tags correct? Currently  openmined/pydp-{branch}-{python-version} Is this consistent across the organization?
```
REPOSITORY                                           TAG                                
openmined/pydp-dev-3.6                               eb3f50ce77c50886051309cdcf7708e7d40ca94a  
openmined/pydp-dev-3.6                               latest   
```
2. Un-comment the last stage to push the images to docker hub and add the credentials to github secrets. 
3. Remove verbose comments and scratch marks in Dockerfile. I left it to assist the reviewer.

Notes: I made changes the the Dockerfile as it cloned the `thirdparty` code which was not pegged to a version or sha. Instead I copy this in by cloning the submodule (same as a local build). l don't see any issues with that but just wanted to mention that incase I have missed something

Dockerfile
chinmayshah99: Yeah. this is what we are aiming at!

.github/workflows/update-docker-image.yml
chinmayshah99: We want to run docker tests one each commit. 
But publishing it to dockerhub is not something we want to do for each commit

Dockerfile
```diff
@@ -45,23 +45,37 @@ WORKDIR ${PROJECT_DIR}
...
+# Either we clone it here as was done but then it should be @e819e03 to match the submodule 
+# and will needed to be updated in keeping with that (a rather tedious way)
+# Get google dp dependency. Rather copy the thrid party deps as it's a cloned submodule
+# COPY ./third_party third_party
 
+# RUN mkdir -p third_party && \
+#     cd third_party && \
+#     git clone https://github.com/google/differential-privacy.git
+
+# Or we just copy it in and ensure the CI/CD clone the repo and the submodule.
```
chinmayshah99: The issue with this is when we do run this in remote environment,  `https://github.com/google/differential-privacy.git` is going to be a submodule which is not going to be cloned. Thus this will break your system. 
You can verify this by simply cloning the project and trying to build the dokcerfile. 
So what we can do here is probably stick to mentioning the sha.

chinmayshah99 (changes requested):

Dockerfile
zizu-research: Please see the build it all passes as I pull all submodules 
In github actions one just need to do:
```
steps:
      - uses: actions/checkout@v2
        with:
          submodules: true
```
"So what we can do here is probably stick to mentioning the sha." 
That was the alternative suggested but it's very hidden I believe it'll break as people **will** forget. 

FYI:
If you want to clone a repo with it's submodules (Which is what I do on my local build) simply use:
```
git clone --recurse-submodules https://github.com/OpenMined/PyDP.git
``` 
This will correctly clone the repo at the sha of upstream as shown in the repo.

BenjaminDev: That is not a problem as we clone the submodules in the github action already 
See the github action file.
```
steps:
      - uses: actions/checkout@v2
        with:
          submodules: true
```

As for the local builds one just needs to clone the submodules using `--recurse-submodules `

FYI:
For running things locally or in any other checkouts one just needs to clone as follows. 
```
git clone --recurse-submodules https://github.com/OpenMined/PyDP.git
```
This will correctly clone the upstream dependencies (`third_party`) at the correct sha. 

"So what we can do here is probably stick to mentioning the sha."
This is an alternative as suggested however, it is a very hidden config an **will** be forgotten and things will break (or not break when they should) as that sha will become out of sync with the one mentioned in the submodule. I'm in favor of letting git handle as much as possible.

However the remaining issues do still remain.
https://github.com/OpenMined/PyDP/pull/176#issuecomment-641100600

chinmayshah99: Oh, I didn't notice the submodule in actions.  
For running locally, we already have scripts that take cares of this stuff.

chinmayshah99: > 1. Are the image names and tags correct? Currently  openmined/pydp-{branch}-{python-version} Is this consistent across the organization?

This looks good. 

> 1. Un-comment the last stage to push the images to docker hub and add the credentials to github secrets.
> 2. Remove verbose comments and scratch marks in Dockerfile. I left it to assist the reviewer.

I'll make these changes while adding the credentials

chinmayshah99 (dismissed):

chinmayshah99:
Hey @BenjaminDev thanks for the PR. Why don't you join us on Slack?

chinmayshah99 (approved):
LGTM!

Issue #192, opened 2020-06-24
Move from tox to pytest


chinmayshah99:
## Description
For python testing:

We don't need tox anymore as most of our stuff is automated using pytest. And in github actions, each tests are run individually in different python containers, thus pytest would suffice our need.

chinmayshah99:
Work can be picked up from #193 since the person can't take it forward https://github.com/OpenMined/PyDP/pull/193#issuecomment-655950659

alejandrosame:
I can take it from here, since it is also raising an error when following Docker instructions:

```
root@050b6b45660c:~/PyDP# make test-all
tox
make: tox: Command not found
make: *** [Makefile:67: test-all] Error 127
```

PR #193, opened 2020-06-24
Move from tox to pytest #192

somiljain7:
#192 
tox is not required anymore


chinmayshah99:
We also need to remove it from github actions and add pytest there

codeboy5:
> We also need to remove it from github actions and add pytest there

Hey @chinmayshah99 I had a doubt, pytest and tox are used for different things, I guess tox is just used to check whether out code works in different python versions as compared to pytest which is used for unit test. Is that correct ?

chinmayshah99:
> Hey @chinmayshah99 I had a doubt, pytest and tox are used for different things, I guess tox is just used to check whether out code works in different python versions as compared to pytest which is used for unit test. Is that correct ?

Yeah they are different. But since we use different containers for different python versions, we no longer need tox. Because in each container, we have a different python version installed.

chinmayshah99 (changes requested):
Can we remove tox tests from github actions too?

somiljain7:
Hey @chinmayshah99 
I was unable to find deciding factor in setup .py
Can you assign someone else for this
Sorry  for disturbance

alejandrosame:
I can take it from here, since it is also raising an error when following Docker instructions:

```
root@050b6b45660c:~/PyDP# make test-all
tox
make: tox: Command not found
make: *** [Makefile:67: test-all] Error 127
```

PR #199, opened 2020-07-22
Added a dockerfile for easy running of examples.

BenjaminDev:
## Description
Following a slack conversation there was a desire for a simple way to run the PyDP examples. It's a minimal docker image with the example dependencies and jupyter installed. 

Summary added to readme:
### Docker alternative
To run the examples in a container:
1. `cd PyDP/examples`
2. `docker build -t pydp-sandbox:local .`
3. `docker run -p 8899:8899 pydp-sandbox:local`
4. Open `displayed url` in a browser. 

## Affected Dependencies
None.

## How has this been tested?
Tested on my local machine following the `examples/readme.md`

## Checklist

### DIFF
--- a/examples/Dockerfile
+++ b/examples/Dockerfile
@@ -1,0 +1,18 @@
+# Pull base image
+ARG PYTHON_VERSION=3.7
+FROM python:${PYTHON_VERSION}-slim-buster
+ENV PROJECT_DIR="${HOME}/examples"
+# Update pip and setuptools and install pipenv
+RUN pip install --upgrade pip setuptools wheel && \
+    pip install pipenv
+WORKDIR ${PROJECT_DIR}
+COPY . .
+# This makes the pipenv's virtual environment in the project dir 
+ENV PIPENV_VENV_IN_PROJECT=true 
+RUN pipenv install python-dp pandas jupyter --skip-lock
+# This `activates` the virtual env
+ENV VIRTUAL_ENV=$PROJECT_DIR/.venv
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+
+EXPOSE 8899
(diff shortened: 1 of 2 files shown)


chinmayshah99:
Can we add this to when we publish it on docker hub?
We can the previous one as dev image and this as sandbox image?

BenjaminDev:
> Can we add this to when we publish it on docker hub?
> We can the previous one as dev image and this as sandbox image?

Sure that is not a problem :grinning: 

However, asking the user to build the image is a one-liner `docker build -t pydp-sandbox:local .`. I guess the question is which is easier for the user:
a)
`docker build -t pydp-sandbox:local .`
`docker run -p 8899:8899 pydp-sandbox:local`
or
b) 
`docker run -p 8899:8899 openmined/pydp-sandbox:latest`

I am in favor of a) as it's done. If we opt for b) then I'd like to try shrink the image some more. As pandas alone uses ~96MB. I'm sure a few flags like `--no-cache` will do the [trick.](https://towardsdatascience.com/how-to-shrink-numpy-scipy-pandas-and-matplotlib-for-your-data-product-4ec8d7e86ee4)  

I'll work on b) in the next few days. For now users can be pointed to solution  a).

chinmayshah99:
People using Sandbox will be directly doing docker pull rather than docker build. So we need to minimize the docker file size!

chinmayshah99:
People using Sandbox will be directly doing docker pull rather than docker build. So we need to minimize the docker file size!

examples/Dockerfile
```diff
@@ -0,0 +1,18 @@
+# Pull base image
```
chinmayshah99, line 1: What base image do you have in mind?

Issue #107, opened 2020-04-19
Documentation for PyDP


chinmayshah99:
We need to start the documentation at `/docs` folder accessible at [here ](pydp.readthedocs.org). 

Before we start the documentation:
1. We need to identify what should be the documentation structure. What should be the template documentation (ex. should we get inspired by how Tensorflow is documented or some other way?)

Once we decide on the template, we should start with the documentation on how to use the library. 

Additionally, #75 also needs to be added in the same documentation structure, in the section contribution guidelines.

chinmayshah99:
Partial work being done at #149

chinmayshah99:
@em-blue We're thinking of using https://www.sphinx-doc.org/en/master/usage/quickstart.html for documentation as it already has in-write support with python code/ pybind11. 

So this requires writing documentation in rst. Was wondering if @OpenMined/writing has a better suggestion?

chinmayshah99:
Done here: 
#207 , #208 , #209 , #210 , #211 , #212

8bitmp3:
@chinmayshah99 Shall we put/duplicate this Issue in GitHub Projects, since there are a few moving parts here? Then we can assign people and track progress in a "better UI"

chinmayshah99:
PyDP team doesn't use GitHub projects yet, but it would probably be a good idea.

PR #206, opened 2020-07-30
Documentation only

BenjaminDev:
## Description
Addresses #107 
Added sphinx documentation [see example](https://benjamindev.github.io/PyDP/readme.html). This means we need to write docstrings in a sphinx compatible way. See `benjaminDev/pydp@distributions` pr for examples. This pr has a workflow that will run on pushes to  `dev` which will build and put the built docs in the `gh-pages` branch. 

## Affected Dependencies
Added dev dependencies: `sphinx` and `sphinx-rtd-theme`.

## How has this been tested?
- Tested on personal fork and it works their.

## Checklist

## Todo:
This pr will has a workflow that will run on pushes to  `dev` which will put the built docs in the `gh-pages` branch. To serve these an Openmined developer must go to settings and activate github pages as shown below.

image


chinmayshah99:
Hey @BenjaminDev ,
We already have readthedocs.org integration over here: https://pydp.readthedocs.io/en/latest/
So we just need to put everything in /docs folder. 

Or is this PR a better solution to that?

BenjaminDev:
hahaha. Okay I missed that. I have never used read the docs. I didn't see a `.readthedocs.yml` file anywhere.

Can you point it to the gh-pages branch? or can readthedocs run the build command?d

I'd say the only issue is somewhere we need to build the docs as it'll get messy if we commit all the built files to github. We'd need to look into what readthedocs offers and whether sphinx can play nicely with sphinx.
Seems it can.
```
sphinx:
  builder: html
  configuration: conf.py
  fail_on_warning: true
```
Where is our .readthedocs.yml file? 

Is this current solution better? I'd say same same. But if other  OpenMined projects are using readthedocs service then let's do the same.

chinmayshah99:
Can we add markdown support?

BenjaminDev:
> Can we add markdown support?

I'm sure we can. which files are in markdown that need to be added? Unless they exist let's leave them to another PR. I'm sure there will be plenty to add.

docs/requirements.txt
```diff
@@ -0,0 +1,2 @@
+python-dp
+sphinx-rtd-theme
```
chinmayshah99, line 2: Can we remove python-dp? or is that essential to the build?

chinmayshah99: Also, is this requirement file necessary?

chinmayshah99:
Also since we have added readthedocs, can we remove the github action?

docs/requirements.txt
BenjaminDev: The way sphinx works is by stripping the docs from the docstrings of the released (installed) package. Yes! this requirements.txt is needed and is used by reathedocs (or the github action proposed) when the docs are built.

BenjaminDev:
> Also since we have added readthedocs, can we remove the github action?

Yeah we should.  I'll do that.

PR #240, opened 2020-08-12
[WIP] Refactoring: Add bounded algorithms

alejandrosame:
## Description
Pushing my proposal towards bounded_algorithms.
Solved merge conflicts adapting ideas from @chinmayshah99's approach.
Updated algorithm test to cover new Python API

## How has this been tested?
`make test` returns successful tests for count and bounded algorithms.
- Provide instructions so we can reproduce.
- List any relevant details for your test configuration.

## Checklist

### DIFF
--- a/pydp/algorithms/algorithm.py
+++ b/pydp/algorithms/algorithm.py
@@ -1,0 +1,41 @@
+from functools import wraps
+
+from .._pydp import _algorithms
+
+
+class Algorithm:
+
+    # Class variables
+    __methods_to_wrap = [
+        "epsilon",
+        "privacy_budget_left",
+        "memory_used",
+        "add_entries",
+        "add_entry",
+        "result",
+        "partial_result",
+        "consume_privacy_budget",
+        "reset",
+        "serialize",
+        "merge",
+        "noise_confidence_interval",
+    ]
+
+    def __init__(self, dtype="int", **kwargs):
+        cpp_class_name = f"{self.__class__.__name__}{self.__map_type_str(dtype)}"
+        class_ = getattr(_algorithms, cpp_class_name)
+
+        self.dtype = dtype
+        self.__algorithm = class_(**kwargs)
+
+        for method in self.__methods_to_wrap:
+            setattr(self, method, getattr(self.__algorithm, f"{method}"))
+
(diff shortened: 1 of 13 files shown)


pydp/algorithms/algorithm.py
```diff
@@ -0,0 +1,41 @@
...
+        class_ = getattr(_algorithms, cpp_class_name)
+
+        self.dtype = dtype
+        self.__algorithm = class_(**kwargs)
+
+        for method in self.__methods_to_wrap:
+            setattr(self, method, getattr(self.__algorithm, f"{method}"))
+
+    @staticmethod
+    def __map_type_str(type):
```
BenjaminDev: typo:
I think `type` should be `dtype` as used in the f-string below.

pydp/algorithms/algorithm.py
```diff
@@ -0,0 +1,41 @@
...
+    # Class variables
+    __methods_to_wrap = [
+        "epsilon",
+        "privacy_budget_left",
+        "memory_used",
+        "add_entries",
+        "add_entry",
+        "result",
+        "partial_result",
+        "consume_privacy_budget",
```
BenjaminDev: Typo:
we need to remove `consume_private_budget` as it has been removed from the builder.

The core lib says it `// Strictly reduces privacy budget, so is safe to make public.` I guess we need to see if it's desired to expose it.

Either way they need to match up.

alejandrosame: The line about consume_private_budget that I deleted previously was simply because it was defined again a couple of lines down (basically I removed it from the  *getters* group as it is not a getter).

In any case, I was just exposing the public methods of algorithm.h by default. Indeed, if it is only meant to be used internally and it makes no sense for the user to call it, we shouldn't expose it.

pydp/algorithms/algorithm.py
alejandrosame: Right! Already fixed.

pydp/algorithms/algorithm.py
alejandrosame: All clear now. I already deleted it both from the Python API and the C++ binding.

PR #252, opened 2020-08-16
Included demo for the restaurant scenario

thefirebanks:
## Description
This is the Python version of Google's differential privacy [restaurant demo](https://github.com/google/differential-privacy/tree/main/examples/java) (originally in Java) 

## Checklist


8bitmp3:
Hi @thefirebanks, thanks for the PyDP demo PR. Quick question - is this is the same differential privacy restaurant demo that @shaistha24 shared in https://github.com/shaistha24/demo/blob/master/PYDP_demo/restaurant_demo.ipynb? I can assist with reviewing the notebook, having done this for TensorFlow and other projects. Cheers! cc @chinmayshah99

chinmayshah99:
Hi @8bitmp3 , Yes this is the same thing @shaistha24 was working on. We keep the notebook for demonstration and playground structure while the python code for the more production version usage of the same. 
Would love have you to be reviewing it!

8bitmp3:
@chinmayshah99 Ok! 

Re: PyDP differential privacy 🥕 carrots demo - will review this asap https://github.com/OpenMined/PyDP/blob/dev/examples/carrots_demo/carrots_demo.ipynb

chinmayshah99:
I'll create an issue for you.

Regards

Chinmay Shah

examples/restaurant_demo/restaurant.py
```diff
@@ -101,84 +106,95 @@ def get_private_counts_per_hour(self, privacy_budget: float) -> dict:
...
         visits = self._hour_visits.copy()
-        visits["Time entered"] = pd.to_datetime(visits["Time entered"]).apply(
-            lambda x: x.strftime(r'%H')).astype(int)
+        visits["Time entered"] = (
+            pd.to_datetime(visits["Time entered"])
+            .apply(lambda x: x.strftime(r"%H"))
+            .astype(int)
+        )
 
         # TODO: Question, should this be the privacy budget? The original code sets epsilon to be log(3)
```
8bitmp3: I assume this PR is still work in progress, so you left some TODOs here 👍

examples/restaurant_demo/restaurant.py
```diff
@@ -98,80 +101,92 @@ def get_private_counts_per_hour(self, privacy_budget: float) -> dict:
...
 
         for day in day_visits["Day"].unique():
-            day_counts[day] = x.result(list(day_visits[day_visits["Day"] == day]["Day"]))
-        
+            day_counts[day] = x.result(
+                list(day_visits[day_visits["Day"] == day]["Day"]))
+
         return day_counts
 
     def get_non_private_sum_revenue(self) -> dict:
```
8bitmp3, line 206: Do you think we should have a docstring for this method?

Also, since PyDP is in early days, it may be best to start using triple quotes/a delimiter (""") instead of a hash (#) for all docstrings. This can potentially help with auto-generating (API) docs.

examples/restaurant_demo/restaurant.py
```diff
@@ -98,80 +101,92 @@ def get_private_counts_per_hour(self, privacy_budget: float) -> dict:
...
 
         for time in visits["Time entered"].unique():
-            private_hours_count[time] = x.result(list(visits[visits["Time entered"] == time]["Time entered"].astype(int)))
-        
+            private_hours_count[time] = x.result(
+                list(visits[visits["Time entered"] == time]["Time entered"].astype(int)))
+
         return private_hours_count
 
     def get_non_private_counts_per_day(self) -> dict:
```
8bitmp3, line 162: Do you think we should have a docstring for this method?

Also, since PyDP is in early days, it may be best to start using triple quotes/a delimiter (""") instead of a hash (#) for all docstrings. This can potentially help with auto-generating (API) docs.

examples/restaurant_demo/restaurant.py
```diff
@@ -1,19 +1,19 @@
+import math
+import statistics as s
+import pydp as dp  # our privacy library
+import pandas as pd
+from collections import defaultdict
 import sys  # isort:skip
 sys.path.append("../pydp")  # isort:skip
```
8bitmp3, line 8: I may have missed this somewhere Do you think having/adding a comprehensive README is/would be a good idea? Example: Restaurant - Differential Privacy by Google: 1) Go: https://github.com/google/differential-privacy/blob/main/examples/go/README.md; 2) Java: https://github.com/google/differential-privacy/blob/main/examples/java/README.md

examples/restaurant_demo/restaurant.py
```diff
@@ -78,18 +78,21 @@ def sum_revenue_per_day_with_preaggregation(self) -> tuple:
     def get_non_private_counts_per_hour(self) -> dict:
```
8bitmp3, line 102: Do you think we should have a docstring for this method?

Also, since PyDP is in early days, it may be best to start using triple quotes/a delimiter (""") instead of a hash (#) for all docstrings. This can potentially help with auto-generating (API) docs.

examples/restaurant_demo/restaurant.py
```diff
@@ -98,80 +101,92 @@ def get_private_counts_per_hour(self, privacy_budget: float) -> dict:
...
             spending = list(visitor_to_spending.values())
-            
+
             day_revenue[day] = int(x.result(spending))
 
         return day_revenue
 
+
 def bound_visits_per_week(df, limit):
     """ Make sure that each visitor only contributes at most 3 visits per week,
```
8bitmp3: Here, you're using triple quotes/a delimiter (""") instead of a hash (#) for the docstring. 
For consistency, I think we should have all docstings with """.
(This can potentially help with auto-generating (API) docs.)

examples/restaurant_demo/restaurant.py
```diff
@@ -40,7 +40,7 @@ class RestaurantStatistics:
     def __init__(self, hours_filename, days_filename, epsilon=LN_3):
```
8bitmp3, line 42: In this constructor/`__init__` function, we should explain via comments what each variable does.

e.g.
```python
...
# Epsilon—the privacy threshold—a number between 0 and 1
self.epsilon = epsilon
```

examples/restaurant_demo/restaurant.py
thefirebanks: That was rather a question that got addressed in Slack! I was confused about how the privacy budget and epsilon were being used

examples/restaurant_demo/restaurant.py
thefirebanks: Do you mean triple quotes for all comments or just doc strings?

examples/restaurant_demo/restaurant.py
thefirebanks: I believe that we will add a README later, just like the carrots demo! right @chinmayshah99 ?

examples/restaurant_demo/restaurant.py
thefirebanks: Gotcha, sounds good!

chinmayshah99:
Yeah

Regards

Chinmay Shah

examples/restaurant_demo/restaurant.py
```diff
@@ -0,0 +1,308 @@
...
+        day_visits = bound_visits_per_week(self._day_visits, COUNT_MAX_CONTRIBUTED_DAYS)
+
+        day_counts = dict()
+
+        # Use the default epsilon value if it is not given as an argument
+        if not epsilon:
+            x = dp.CountInt(self._epsilon)
+        else:
+            x = dp.CountInt(epsilon)
+
```
replomancer, line 197: Here before calling `x.result(...)` we need to set l0 sensitivity on x. Otherwise the default of 1 will be used and that's not enough (the results will not be DP). This is the first case where the default sensitivity is incorrect, it needs to be set in most cases.

examples/restaurant_demo/restaurant.py
```diff
@@ -0,0 +1,308 @@
...
+        day_visits = bound_visits_per_week(self._day_visits, SUM_MAX_CONTRIBUTED_DAYS)
+
+        day_revenue = dict()
+
+        # Use the default epsilon value if it is not given as an argument
+        if not epsilon:
+            x = dp.BoundedSum(self._epsilon, MIN_EUROS_SPENT, MAX_EUROS_SPENT_1)
+        else:
+            x = dp.BoundedSum(epsilon, MIN_EUROS_SPENT, MAX_EUROS_SPENT_1)
+
```
replomancer, line 242: Here also l0 sensitivity needs to be set (to SUM_MAX_CONTRIBUTED_DAYS).

It's a [method call](https://github.com/google/differential-privacy/blob/9923ad4ee1b84a7002085e50345fcc05f2b21bcb/examples/java/SumRevenuePerDay.java#L100) in Java:
```java
              .maxPartitionsContributed(MAX_CONTRIBUTED_DAYS)
```

examples/restaurant_demo/restaurant.py
```diff
@@ -0,0 +1,308 @@
...
+        day_visits = bound_visits_per_week(self._day_visits, SUM_MAX_CONTRIBUTED_DAYS)
+
+        day_revenue = dict()
+
+        # Use the default epsilon value if it is not given as an argument
+        if not epsilon:
+            x = dp.BoundedSum(self._epsilon, MIN_EUROS_SPENT, MAX_EUROS_SPENT_2)
+        else:
+            x = dp.BoundedSum(epsilon, MIN_EUROS_SPENT, MAX_EUROS_SPENT_2)
+
```
replomancer, line 283: Also here.

replomancer (changes requested):

examples/restaurant_demo/restaurant.py
thefirebanks: Added

examples/restaurant_demo/restaurant.py
thefirebanks: Added

examples/restaurant_demo/restaurant.py
thefirebanks: Added

chinmayshah99:
Can we use `add_entries()` function? and then call the `result()` function? More on it here: https://pydp.readthedocs.io/en/latest/pydp.html#pydp.algorithms.laplacian.BoundedMean.add_entries

Also, if we want to keep it in the current form, we may need to move to the `quick_result()` function since in the 0.7 release, we moved `result()` to `quick_result()`

chinmayshah99 (changes requested):
Can we use add_entries() function? and then call the result() function? More on it here: https://pydp.readthedocs.io/en/latest/pydp.html#pydp.algorithms.laplacian.BoundedMean.add_entries

Also, if we want to keep it in the current form, we may need to move to the quick_result() function since in the 0.7 release, we moved result() to quick_result()

chinmayshah99 (dismissed):
LGTM

examples/restaurant_demo/restaurant.py
```diff
@@ -0,0 +1,341 @@
+import pydp as dp
+from pydp.algorithms.laplacian import BoundedSum, Count
+
+import math
+import statistics as s
+import pandas as pd
+from collections import defaultdict
+
+# An hour when visitors start entering the restaurant (900 represents 9:00 AM)
```
8bitmp3: Suggestion:

```diff
- # An hour when visitors start entering the restaurant (900 represents 9:00 AM)
+ # Assumptions:
+ 
+ # Time when visitors start entering the restaurant (900 represents 9:00 AM)
```

Proposing to add "assumptions" here explicitly, since these are hard-coded variables (as integers) we're using in the entire exercise

examples/restaurant_demo/restaurant.py
```diff
@@ -0,0 +1,341 @@
...
+OPENING_HOUR = 900
+
+# An hour when visitors stop entering the restaurant (2000 represents 20:00 PM)
```
8bitmp3:
```diff
- # An hour when visitors stop entering the restaurant (2000 represents 20:00 PM)
+ # Time when visitors finish entering the restaurant (2000 represents 20:00 PM)
```

examples/restaurant_demo/restaurant.py
```diff
@@ -0,0 +1,341 @@
...
+CLOSING_HOUR = 2000
+
+# Range of valid work hours when a visitor can enter the restaurant.
```
8bitmp3:
```diff
- # Range of valid work hours when a visitor can enter the restaurant.
+ # A range—a Python list—of working hours when visitors can come to the restaurant
```

Perhaps we can add a small explanation of why `CLOSING_HOUR + 1` (the `+ 1` bit)?

e.g.

```diff
+ Visitors can stay for 1 hour after the restaurant closes doors to new visitors (`CLOSING_HOUR + 1`)
```

examples/restaurant_demo/restaurant.py
```diff
@@ -0,0 +1,341 @@
...
+VALID_HOURS = list(range(OPENING_HOUR, CLOSING_HOUR + 1))
+
+# For how many hours visitors can enter the restaurant
```
8bitmp3: I'm not 100% sure but _maybe_:

```diff
- # For how many hours visitors can enter the restaurant
+ # The number of total hours during which visitors can enter the restaurants
```

(unless you mean "can stay" instead of "can enter")

WDYT?

examples/restaurant_demo/restaurant.py
```diff
@@ -0,0 +1,341 @@
...
+NUM_OF_WORK_HOURS = CLOSING_HOUR - OPENING_HOUR + 1
+
+# Number of visit days contributed by a single visitor will be limited to 3. All exceeding visits will be discarded.
```
8bitmp3: Perhaps this:

```diff
- # Number of visit days contributed by a single visitor will be limited to 3. All exceeding visits will be discarded.
+ # Cap the maximum number of visiting days at 3 per each visitor (any number above will not be taken into account)
```

examples/restaurant_demo/restaurant.py
```diff
@@ -0,0 +1,341 @@
...
+COUNT_MAX_CONTRIBUTED_DAYS = 3
+
+# Number of visit days contributed by a single visitor will be limited to 4. All exceeding visits will be discarded.
```
8bitmp3: Question - can you help me understand what the difference between COUNT and SUM in the context of CONTRIBUTED DAYS is? 

I'm thinking:

```diff
- # Number of visit days contributed by a single visitor will be limited to 4. All exceeding visits will be discarded.
+ # Cap the maximum number of visiting days at 4 per each visitor (any number above will not be taken into account)
```

But we should differentiate the max. at 3 and 4 here.

examples/restaurant_demo/restaurant.py
```diff
@@ -0,0 +1,341 @@
...
+SUM_MAX_CONTRIBUTED_DAYS = 4
+
+# Minimum amount of money we expect a visitor to spend on a single visit.
```
8bitmp3:
```diff
- # Minimum amount of money we expect a visitor to spend on a single visit.
+ # Expected minimum amount of money (in Euros) to be spent by a visitor per a single visit
```

examples/restaurant_demo/restaurant.py
```diff
@@ -0,0 +1,341 @@
...
+MIN_EUROS_SPENT = 0
+
+# Maximum amount of money we expect a visitor to spend on a single visit.
```
8bitmp3:
```diff
- # Maximum amount of money we expect a visitor to spend on a single visit.
+ # Expected maximum amount of money (in Euros) to be spent by a visitor per a single visit
```

(I recommend to not use "we" if you can)

examples/restaurant_demo/restaurant.py
```diff
@@ -0,0 +1,341 @@
...
+MAX_EUROS_SPENT_1 = 50
+MAX_EUROS_SPENT_2 = 65
+
+LN_3 = math.log(3)
+
+
```
8bitmp3, line 35: Provide a small definition for the `RestaurantStatistics`

e.g.

```diff
+ # Basic restaurant statistics class
class RestaurantStatistics:
...
```

It's self-evident but helpful

examples/restaurant_demo/restaurant.py
```diff
@@ -0,0 +1,341 @@
...
+class RestaurantStatistics:
+    def __init__(self, hours_filename, days_filename, epsilon=LN_3):
+
+        # Store the name of the csvs for daily and weekly restaurant data
```
8bitmp3: CSV _files_ or .csv _files_ - to be more explicit

```diff
-         # Store the name of the csvs for daily and weekly restaurant data
+        # Store the filenames (in csv) with daily and weekly restaurant data
```

examples/restaurant_demo/restaurant.py
```diff
@@ -0,0 +1,341 @@
...
+
+        # The privacy threshold, a number between 0 and 1
+        self._epsilon = epsilon
+
+        # The hourly and weekly data as pandas DataFrames
+        self._hour_visits = pd.read_csv(self.hours_filename, sep=",")
+        self._day_visits = pd.read_csv(self.days_filename, sep=",")
+
+    def count_visits_per_hour(self) -> tuple:
+        """Compute raw count of visits per hour of day and return two dictionaries
```
8bitmp3:
```diff
-        """Compute raw count of visits per hour of day and return two dictionaries 
+        """Compute the number of visits per hour of day and return two dictionaries  
...
```

examples/restaurant_demo/restaurant.py
```diff
@@ -0,0 +1,341 @@
...
+        
+        The first dictionary is the count calculation without any differential privacy, while the second one uses the PyDP library for a private calculation
+        """
+        non_private_counts = self.get_non_private_counts_per_hour()
+        private_counts = self.get_private_counts_per_hour()
+
+        return non_private_counts, private_counts
+
+    def count_visits_per_day(self) -> tuple:
+        """Compute raw count of visits per day of week and return two dictionaries
```
8bitmp3:
```diff
-        """Compute raw count of visits per day of week and return two dictionaries 
+        """Compute the number of visits per hour of day and return two dictionaries  
...
```

examples/restaurant_demo/restaurant.py
```diff
@@ -0,0 +1,341 @@
...
+        The first dictionary is the count calculation without any differential privacy, 
+        while the second one uses the PyDP library for a private calculation
+        """
+        non_private_counts = self.get_non_private_counts_per_day()
+        private_counts = self.get_private_counts_per_day()
+
+        return non_private_counts, private_counts
+
+    def sum_revenue_per_day(self) -> tuple:
+        """Compute revenue per day for the whole week.
```
8bitmp3:
```diff
-         """Compute revenue per day for the whole week. 
+        """Compute the restaurant's daily revenue for the whole week. 
```

examples/restaurant_demo/restaurant.py
```diff
@@ -0,0 +1,341 @@
...
+        The first dictionary is the count calculation without any differential privacy, 
+        while the second one uses the PyDP library for a private calculation
+        """
+        non_private_sum = self.get_non_private_sum_revenue()
+        private_sum = self.get_private_sum_revenue_with_preaggregation()
+
+        return non_private_sum, private_sum
+
+    def get_non_private_counts_per_hour(self) -> dict:
+        """Compute the number of visits per hour without any differential privacy. Return a dictionary mapping hours to number of visits"""
```
8bitmp3: Here and elsewhere, it would look better if you lint it - cap at 80 characters per line

examples/restaurant_demo/restaurant.py
```diff
@@ -0,0 +1,341 @@
...
+
+if __name__ == "__main__":
+    r = RestaurantStatistics(
+        hours_filename="day_data.csv", days_filename="week_data.csv"
+    )
+
+    np_count_hour, p_count_hour = r.count_visits_per_hour()
+    print("-----------------------------------")
+    print("Visits per hour:")
+    print("Non-private:", np_count_hour)
```
8bitmp3: Suggestion:

Maybe instead of saying "Non-private:" and "Private:" in print statements, we can say "Without differential privacy:" and "With differential privacy:", respectively? ("Private" can have many meanings.)

8bitmp3 (changes requested):
Awesome work @thefirebanks ! The attention to detail is 👍 🙌 🚀

I've taken a look and provided some suggestions in `restaurant.py`. Feel free to use them as guidelines and, of course, you can ignore them - I'm jus trying to help.

Some of the suggestions:

- Maybe we should add "assumptions" from the beginning explicitly, since these are hard-coded variables (as integers) we're using in the entire exercise. It's not a notebook, but users should be reading the code from the beginning (import libraries, hard-code some variables, etc)

```diff
- # An hour when visitors start entering the restaurant (900 represents 9:00 AM)
+ # Assumptions:
+ 
+ # Time when visitors start entering the restaurant (900 represents 9:00 AM)
```

- Some minor refactoring, such as:

```diff
- # Range of valid work hours when a visitor can enter the restaurant.
+ # A range—a Python list—of working hours when visitors can come to the restaurant
```

```diff
-        """Calculates revenue per day in the whole week while preagreggating the spending of each visitor before calculating the BoundedSum with PyDP.
+        """Calculates the restaurant's daily revenue for the whole week, while 
+        pre-agreggating each visitor's spending before calculating the 
+        BoundedSum with PyDP.
```

(note the 80-character cap per line)

- Perhaps we can add a small explanation of why `CLOSING_HOUR + 1` (the `+ 1` bit)?

e.g.

```diff
+ Visitors can stay for 1 hour after the restaurant closes doors to new visitors (`CLOSING_HOUR + 1`)
```

- Provide a small definition for the `RestaurantStatistics` - it's self-evident but helpful

e.g.

```diff
+ # Basic restaurant statistics class
class RestaurantStatistics:
...
```

-Question - can you help me understand what the difference between COUNT and SUM in the context of CONTRIBUTED DAYS is? We should differentiate the max. at 3 and 4 here, since they currently have similar explanations in comments.

e.g. 
```diff
- # Number of visit days contributed by a single visitor will be limited to 4. All exceeding visits will be discarded.
+ # Cap the maximum number of visiting days at 4 per each visitor (any number above will not be taken into account)
```
?

- Maybe instead of saying "Non-private:" and "Private:" in print statements, we can say "Without differential privacy:" and "With differential privacy:", respectively? ("Private" can have many meanings.)

(For more, see comments).

Thank you!

examples/restaurant_demo/restaurant.py
thefirebanks: I realized we aren't really using the variable so I took it out!

examples/restaurant_demo/restaurant.py
thefirebanks: Done!

examples/restaurant_demo/restaurant.py
thefirebanks: Done

examples/restaurant_demo/restaurant.py
thefirebanks: Done

examples/restaurant_demo/restaurant.py
thefirebanks: Done

examples/restaurant_demo/restaurant.py
thefirebanks: There's not a difference besides replicating the code from the original demo, they used different variables so I wrote both of them here as well.

examples/restaurant_demo/restaurant.py
thefirebanks: Done!

examples/restaurant_demo/restaurant.py
thefirebanks: Good point, thanks :)

examples/restaurant_demo/restaurant.py
thefirebanks: I'll add some docs here instead!

examples/restaurant_demo/restaurant.py
thefirebanks: Done

examples/restaurant_demo/restaurant.py
thefirebanks: Done

examples/restaurant_demo/restaurant.py
thefirebanks: Done

examples/restaurant_demo/restaurant.py
thefirebanks: Done

examples/restaurant_demo/restaurant.py
thefirebanks: Gotcha!

examples/restaurant_demo/restaurant.py
thefirebanks: You are absolutely right, done!

8bitmp3:
Thank you @thefirebanks!

@thefirebanks @chinmayshah99 - what do you think about using a Python docstring convention, such as the Google one? https://google.github.io/styleguide/pyguide.html#244-decision 

Currently, in each method we have enough material for:
- The main part that summarizes the behaviour (`""" Does a thing...`)
- The return values (`... Returns: a thing .... a dict/int/tuple"""`)

The users may notice that currently we don't follow a particular style (e.g. `"""Do a thing. Return a thing"""` + `"""Does a thing. Returns a thing"""` + `"""Does a thing and return something. That thing is a thing"""`).

What we can aim for is something like this example from the Google's style guide for Python:

image

If you're OK with using this kind of doc string style, I can help and speed things up by refactoring `restaurant.py` (Auto Docstring generator for Python for VSCode: `https://open-vsx.org/extension/njpwerner/autodocstring` is a good extension). 

Then, I can submit a PR into @thefirebanks 's repo. 

WDYT?

chinmayshah99:
> If you're OK with using this kind of doc string style, I can help and speed things up by refactoring `restaurant.py` (Auto Docstring generator for Python for VSCode: `https://open-vsx.org/extension/njpwerner/autodocstring` is a good extension).

This looks super cool and something we could use throughout our repo. If it's possible, can you send a PR applying it across the repo?
I'll be happy to chip in if something needs to be added.

chinmayshah99 (approved):

PR #264, opened 2020-08-22
Allow int64 as dtype

### LINKED ISSUE #174: PyDP crashes when used with large lists
## Description

This program:
```python
import pydp

example_list = [5 for _ in range(1000000000)]
print("The previous instruction did not crash the script.", flush=True)
print(pydp.BoundedMean(1.0, 0, 10).result(example_list))
```
crashes with
```
The previous instruction did not crash the script.
Killed
```
output. It's the same when I replace `BoundedMean` with `BoundedSum` or `Count`. I suspect that most or all algorithms are crashing like this. It works fine when `example_list` is smaller.
 
## How to Reproduce

Running the above code will probably produce same results on other machines. If not, try with a longer list. Actually the list creation can also crash the script for you, but that's still relevant: the library expects a list input (iterable is not enough), so working with more data gets challenging pretty fast.

## Expected Behavior

I would expect PyDP to handle big input or provide API to add data in batches e.g. the C++ library has `AddEntries` method which can be called multiple times before computing result. 
PyDP could expose `AddEntries` or partition input data before calling `AddEntries` in a loop when `result` is called. 

## System Information
 - OS: Ubuntu
 - OS Version: 20.04
 - Language Version: Python 3.8

## Additional Context

Python's limit for a list size is not an issue here. Here's how you can see the list size limit on your system:
```python
>>> import sys
>>> sys.maxsize
9223372036854775807
```
so I'm nowhere near size limits in my example.

alejandrosame:
## Description
This PR fixes #174. In that issue @chinmayshah99 exposes an issue when trying to compute the mean of a big list, where an overflow causes the result to become 0. By introducing `long long` as an underlying type, we enable the correct computation of these big lists.  

## Affected Dependencies
No dependencies affected.

## How has this been tested?
- Added a new test case `test_bounded_mean_long_long`.
- `make test` executes successfully.
- After compiling and installing on a Raspberry Pi 4 (OS Raspbian [Buster 10] 32 bit), `make test` also executes successfully.

## Checklist

### DIFF
--- a/src/bindings/PyDP/pydp_lib/algorithm_builder.hpp
+++ b/src/bindings/PyDP/pydp_lib/algorithm_builder.hpp
@@ -92,8 +92,8 @@ class AlgorithmBuilder {
   std::map<std::type_index, std::string> type_to_name = {
       {typeid(double), "Double"},
       {typeid(int), "Int"},
-      {typeid(int64_t), "Int"},
-      {typeid(long long), "LongLong"}};
+      {typeid(int64_t), "Int64"},
+  };
   std::map<std::type_index, std::string> algorithm_to_name = {
       {typeid(dp::BoundedMean<T>), "BoundedMean"},
       {typeid(dp::BoundedSum<T>), "BoundedSum"},
--- a/src/bindings/PyDP/pydp_lib/algorithm_builder.hpp
+++ b/src/bindings/PyDP/pydp_lib/algorithm_builder.hpp
@@ -90,7 +90,10 @@ class AlgorithmBuilder {
   }
 
   std::map<std::type_index, std::string> type_to_name = {
-      {typeid(double), "Double"}, {typeid(int), "Int"}, {typeid(int64_t), "Int"}};
+      {typeid(double), "Double"},
+      {typeid(int), "Int"},
+      {typeid(int64_t), "Int"},
+      {typeid(long long), "LongLong"}};
   std::map<std::type_index, std::string> algorithm_to_name = {
       {typeid(dp::BoundedMean<T>), "BoundedMean"},
       {typeid(dp::BoundedSum<T>), "BoundedSum"},
--- a/pydp/algorithms/_algorithm.py
+++ b/pydp/algorithms/_algorithm.py
@@ -27,6 +27,8 @@ class MetaAlgorithm:
             return "Int"
         elif dtype == "float":
             return "Double"
+        elif dtype == "long long":
+            return "LongLong"
         else:
             raise RuntimeError("dtype: {} is not supported".format(dtype))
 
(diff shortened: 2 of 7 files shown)


src/bindings/PyDP/pydp_lib/algorithm_builder.hpp
```diff
@@ -90,7 +90,10 @@ class AlgorithmBuilder {
   }
 
   std::map<std::type_index, std::string> type_to_name = {
-      {typeid(double), "Double"}, {typeid(int), "Int"}, {typeid(int64_t), "Int"}};
+      {typeid(double), "Double"},
+      {typeid(int), "Int"},
+      {typeid(int64_t), "Int"},
```
chinmayshah99: I think `int64_t` and `int` should be a different type when initializing python type. Also, I just went through https://stackoverflow.com/questions/13604137/definition-of-int64-t and I'm not sure if we need `long long` in this case. @BenjaminDev thoughts?

alejandrosame: So, for example, we could take out `long long`and convert it to `int64`. I like that approach better, tbh.  See also https://stackoverflow.com/a/4166592.

chinmayshah99:
Should we wait for Benjamin's opinion on this?

Regards

Chinmay Shah

alejandrosame:
I think we can still wait a day or two.

src/bindings/PyDP/pydp_lib/algorithm_builder.hpp
chinmayshah99: Should we offer both int64 and uint64? 
Or is offering uint64 a violation of DP properties? @OpenMined/dp-research ?

chinmayshah99:
Let's move to int64. But we need to test first if the system supports 64-bit numbers.

alejandrosame:
> Let's move to int64. But we need to test first if the system supports 64-bit numbers.

I think that test is not needed, check https://stackoverflow.com/questions/22363503/will-a-long-integer-work-on-a-32-bit-system, it's mostly a compilation matter. 

AFAIK, `pip install python-dp` currently wouldn't work on 32 bit systems because we always compile Google DP for 64 bit systems. If we want to officially support 32 bit systems, we need to add builds for it in order to streamline testing. I didn't think about it before, but it might be interesting to at least bring support for ARM 32 bit for edge computing, educational purposes, etc (I'm thinking about RPis). 

I have a RPi around, so I'll try to set it up to at least manually test these changes on a 32 bit system.

chinmayshah99:
If the PR is completed, we can remove WIP from the title

alejandrosame:
The manual test on a Raspberry Pi 4 (32 bit OS) was also successful. Taking away the WIP tag.

chinmayshah99 (approved):

PR #267, opened 2020-08-23
adds Docker build on PR 

### LINKED ISSUE #260: Add Docerkfile tests in Github Action for every PR
## Feature Description
Our current tests on each commit are only checked on Linux. But sometimes things are broken for Dockerfile in the process. We need to add it as parts of tests for each commit.

Vishwaak:
## Description
Closes #260, As requested by @chinmayshah99  every PR to dev branch triggers a docker build to check if the PR breaks the docker image or not.

## Affected Dependencies
black version is hardcoded to 19.10b0

## How has this been tested?
- Checked on test PR

## Checklist

### DIFF
--- a/.github/workflows/docker-test.yml
+++ b/.github/workflows/docker-test.yml
@@ -1,0 +1,38 @@
+on:
+    push:
+      branches:
+      - dev
+    pull_request:
+      types: [opened, synchronize, reopened]
+
+name: Build and Test Docker Image
+jobs:
+  publish:
+    name: Build and publish
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.8]
+    env:
+      IMAGE: openmined/pydp
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+     
+      - name: Show python version and build image name  
+        id: build_name
+        run: |
+          python -c "import sys; print(sys.version)"
+          echo "::set-output name=image_name::${IMAGE}-${GITHUB_REF##*/}-${{matrix.python-version}}"
+
+      - name: Build images
+        run: docker build -t ${{steps.build_name.outputs.image_name}} --build-arg PYTHON_VERSION=${{matrix.python-version}} -f Dockerfile .
+
+   
+
--- a/Pipfile
+++ b/Pipfile
@@ -7,7 +7,7 @@ verify_ssl = true
 setuptools = "*"
 bumpversion = "==0.5.3"
 pytest = "*"
-black = "*"
+black = "==19.10b0"
 twine = "*"
 sphinx = "*"
 sphinx-rtd-theme = "*"


.github/workflows/docker-test.yml
```diff
@@ -0,0 +1,38 @@
...
+      types: [opened, synchronize, reopened]
+
+name: Update and Test Docker Image
+jobs:
+  publish:
+    name: Build and publish
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8]
```
chinmayshah99: I don't think we need to check for all python versions. One should be enough. Maybe stick to 3.8?

Vishwaak: I thought since we have release configured for 3.6,3.7,3.8 so I thought it would be better to check for them too. Alright, will get it patched up.

chinmayshah99 (approved):
LGTM

.github/workflows/docker-test.yml
chinmayshah99: We are anyways checking it on CLI. The whole idea of this is that during each PR, it shouldn't break the docker file.

BenjaminDev (commented):
This PR only builds the docker image but does not test that the built image is actually working. I'm not sure if this is desired but I like to put something like 
```
   - name: Run tests
         run: | 
           docker run ${{steps.build_name.outputs.image_name}}:latest /bin/bash -c  "pipenv install --dev && pipenv run pytest ./tests"
```
Which runs the test suit against the built library in side the docker image. That way we know it's all working. 
Other than that it looks good to me.

alejandrosame (approved):
LGTM. I agree with @BenjaminDev, if we do not add it already I would make an issue to make sure that the container is tested down the line.

chinmayshah99:
@BenjaminDev @alejandrosame The idea of #260 and this PR is to avoid issues during release as faced here: https://github.com/OpenMined/PyDP/runs/996868032?check_suite_focus=true

alejandrosame:
@chinmayshah99 If we are publishing official images, then it is even more relevant to test that the built container passes the tests. For example, it is possible to change the Dockerfile in a way that builds without complaining but fails at execution. The more extreme case is by simply not adding the package, but it could be due to more subtle changes like ENV variables.

This PR can move on as it is, but I do see the value on testing the container built!

chinmayshah99:
> This PR can move on as it is, but I do see the value on testing the container built!

Just to confirm, testing each container for all PRs?

alejandrosame:
I think it has to be for PRs that aim to be published, so this maybe only applies to specific branches. Eventually, we should aim to cover as much as it's possible anything that is going to be published.

chinmayshah99:
In that case, feel free to create another issue. Let's merge this PR in the
meantime.

Regards

Chinmay Shah

chinmayshah99:
hi @Vishwaak, there seems like a mismatch between the python versions. In one step, we are saying we are using python 38, but then we are using py37 underneath the hood. More on it here: https://github.com/OpenMined/PyDP/pull/278/checks?check_run_id=1042574486#step:5:3
Can this typo be fixed?

Vishwaak:
Hello, @chinmayshah99 that arg value sets the default value of the python version, and uses it if not specified during run time. But it's not the case here.

chinmayshah99:
In that case, just mention in the job name that we are using py37.

Vishwaak:
Wouldn't that contradict the fact that we are using py3.8, to build the image?

PR #271, opened 2020-08-26
Fix core crashes due to bound params

### LINKED ISSUE #268: Passing float("NaN"), float("inf") or float("-inf") as lower/upper bounds crashes the interpreter
## Description
When I instantiate some algorithms with `NaN` as a bound, Python crashes:
```python
In [1]: import pydp.algorithms.laplacian as dp

In [2]: x = dp.BoundedMean(0.1, 0.0, float("NaN"), dtype="float")
2020-08-24 20:40:05  FATAL  distributions.cc : 81 : Check failed: lambda >= 0 (nan vs. 0) 
2020-08-24 20:40:05  FATAL  distributions.cc : 81 : Check failed: lambda >= 0 (nan vs. 0) 
Aborted (core dumped)
``` 
I see the same result with `dp.BoundedMean(0.1, float("NaN"), 10.0, dtype="float")` (lower bound set to `NaN`)
and also with `BoundedSum`.

## How to Reproduce
See above example.

## Expected Behavior
When I try to pass `NaN` as epsilon I get an exception without a crash. It should be the same with lower and upper bounds.
 
## System Information
 - OS: Ubuntu (Docker image)
 - OS Version: (Docker image)
 - Language Version: 3.7.7 (Docker image)

alejandrosame:
## Description
Fixes #268. Given that the underlying C++ code does not accept NaN values and complains that "inf" values are too high, the approach followed by this was to simply raise ValueError exceptions when bound params are set to "NaN", "-inf" or "inf".

## Affected Dependencies
No dependencies changed.

## How has this been tested?
- Introduced new tests to cover this case.
- `make test` runs successfully.

## Checklist

### DIFF
--- a/pydp/algorithms/_algorithm.py
+++ b/pydp/algorithms/_algorithm.py
@@ -40,7 +40,7 @@ class MetaAlgorithm:
         elif dtype == "float":
             return "Double"
         else:
-            raise RuntimeError("dtype: {} is not supported".format(dtype))
+            raise ValueError("dtype '{}' is not supported.".format(dtype))
 
     @property
     def epsilon(self) -> float:
--- a/pydp/algorithms/_algorithm.py
+++ b/pydp/algorithms/_algorithm.py
@@ -1,3 +1,5 @@
+import math
+
 from .._pydp import _algorithms
 
 from typing import Union, List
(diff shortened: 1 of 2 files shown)


pydp/algorithms/_algorithm.py
```diff
@@ -21,14 +26,21 @@ def __init__(self, **kwargs):
...
+
     @staticmethod
     def __map_dtype_str(dtype: str):
         if dtype == "int":
             return "Int"
         elif dtype == "float":
             return "Double"
         else:
-            raise RuntimeError("dtype: {} is not supported".format(dtype))
+            raise ValueError("dtype '{}' is not supported.".format(dtype))
```
chinmayshah99, line 43: Shouldn't this be TypeError?

alejandrosame: TypeError is used when an unexpected Python type is received. In this case the type of "dtype" is str and we do not accept values outside the main cases ("int" and "float" for the moment).

chinmayshah99 (approved):

PR #307, opened 2020-10-10
Laplace Demo

### LINKED ISSUE #209: Laplacian mechanism introduction
## Who?
Developer who's trying to understand what does it mean when you select laplacian

## What?
Explanation of what choosing Laplacian mechanism for DP means and underlying assumption as well as recommendations on how noise is added and what type of data is desirable. 

## Additional Context
recommend working actively with research team

NateSolon:
## Description
A jupyter notebook demonstrating and explaining the Laplacian mechanism, the most common way of adding noise in differential privacy.

## Affected Dependencies
None

## How has this been tested?
Requested and made changes based on feedback from members of the PyDP and Writing teams.

## Checklist
- [ x] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [ x] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -70,6 +70,7 @@ jobs:
          pipenv run python -m pip install --force-reinstall $wheel_path
     - name: Run tests
       run: |
+        pipenv install matplotlib
         pipenv run coverage run -m pytest tests
     - name: Check Python code coverage
       if: runner.os == 'Linux' # Coverage will be the same on all systems so only running the check on Linux
@@ -80,7 +81,6 @@ jobs:
     - name: Mypy test
       run: |
         pipenv install mypy
-        pipenv install matplotlib
         pipenv run mypy tests/ pydp/
         pipenv run mypy examples/carrots_demo examples/restaurant_demo examples/laplace_demo
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -80,7 +80,7 @@ jobs:
     - name: Mypy test
       run: |
         pipenv install mypy
-        pipenv install matplotlib --ignore-pipfile --skip-lock
+        pipenv install matplotlib
         pipenv run mypy tests/ pydp/
         pipenv run mypy examples/carrots_demo examples/restaurant_demo examples/laplace_demo
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -80,6 +80,7 @@ jobs:
     - name: Mypy test
       run: |
         pipenv install mypy
+        pipenv install matplotlib --ignore-pipfile --skip-lock
         pipenv run mypy tests/ pydp/
         pipenv run mypy examples/carrots_demo examples/restaurant_demo examples/laplace_demo
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -81,7 +81,7 @@ jobs:
       run: |
         pipenv install mypy
         pipenv run mypy tests/ pydp/
-        pipenv run mypy examples/carrots_demo examples/restaurant_demo
+        pipenv run mypy examples/carrots_demo examples/restaurant_demo examples/laplace_demo
 
     # Need to see how to run the c++ code coverage
     # - name: Check C++ code coverage
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -82,7 +82,7 @@ jobs:
       run: |
         pipenv install mypy
         pipenv run mypy tests/ pydp/
-        pipenv run mypy examples/carrots_demo examples/restaurant_demo examples/laplace_demo
+        pipenv run mypy examples/carrots_demo examples/restaurant_demo
 
     # Need to see how to run the c++ code coverage
     # - name: Check C++ code coverage
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,3 @@
-matplotlib
 python-dp
 sphinx-rtd-theme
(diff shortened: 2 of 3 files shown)


chinmayshah99:
Hi @NateSolon ,
You will need to add the package dependencies in test.yml file similar to done here: https://github.com/OpenMined/PyDP/blob/dev/.github/workflows/tests.yml#L57

NateSolon:
Sorry, I don't understand. Where/how do I add the package dependencies?

8bitmp3 (changes requested):
Hi @NateSolon - awesome stuff! 

May I recommend using a notebook linter that helps you check if you're following certain documentation style guides, such as the [Google style guide](https://developers.google.com/style) (it's not a rule book but there's some good advice with a goal to help the end users of all abilities and backgrounds). 

Keep in mind that we should strive to make the language accessible and inclusive (https://developers.google.com/style/inclusive-documentation). Also, you should use the second person (`you` instead of `we`) in order to directly speak to the user (https://developers.google.com/style/person). Again, these are just guidelines.

About linting:

> The easiest way to scale reviews is to let the machine do it. Every project has recurring issues that pop up in reviews, and style questions are often best settled with a style guide...

(from this [blog post](https://blog.tensorflow.org/2020/10/how-tensorflow-docs-uses-juypter-notebooks.html) by @lamberta)

1. In your local (virtual) env, you can install the TensorFlow team's notebook tools:

`python3 -m pip install -U git+https://github.com/tensorflow/docs`

2. Then use the `nblint` linter with the `--styles=google` flag:

`python3 -m tensorflow_docs.tools.nblint  --styles=google your_notebook.ipynb`

This doesn't highlight everything, but it's good. Check out this [README](https://github.com/tensorflow/docs/blob/master/tools/tensorflow_docs/tools/README.md#nblint) for more options.

NateSolon:
@8bitmp3 Thanks! I was able to run the linter locally and like you said, it suggests changing "we" to "you". I can go ahead and make that change.

It seems what's causing my PR to fail is that the matplotlib import isn't working for some versions of python. Do you know how I can address that?

NateSolon:
@8bitmp3 Hi, I switched "we" to "you" in the notebook as discussed.

@chinmayshah99 Apologies if I'm missing something obvious here, but it looks like a lot of tests are failing because the matplotlib import isn't working. Do I need to do something in my PR to address that?

8bitmp3 (changes requested):
After a quick look, here's my 2 cents:

I think you should gently introduce the readers to PyDP in one short (but not too short) paragraph before the first paragraph. Otherwise, you're kind of starting in the middle and those who aren't familiar with the the concept of "noise" may assume you skipped something. A good short intro is present in most good notebook tutorials. Check out TensorFlow's notebooks, for example - they have a dedicated docs team and a big community.

You could _rephrase_ this part from the README:

PyDP is a library that "provides a set of ε-differentially private algorithms, which can be used to produce aggregate statistics over numeric data sets containing private or sensitive information." It helps "control the privacy guarantee and accuracy of your model written in Python.

PyDP "features differentially private algorithms", such as ... (name a few.)

Also, I think you should a quick explanation to what this "noise" thing is in after "...adding noise". Also, the word "attacker" assumes someone's after the data. Since most stuff we have nowadays doesn't use diff privacy (I assume), let's call them "A typical user" (or if you have a better idea - let's try something else).

It may also be useful to have an intro similar to the Carrots demo https://github.com/NateSolon/PyDP/blob/dev/examples/carrots_demo/carrots_demo.ipynb (the stuff before the Zoo Animals section). I will lint and edit `carrots_demo.ipynb` to improve it if @chinmayshah99 thinks it's a good idea. I've already spotted a few things (e.g. We have section `1.1` but nothing else is numbered, so we can remove it).

Update:

I'd also recommend the ideas from https://github.com/numpy/numpy-tutorials/blob/master/content/tutorial-style-guide.ipynb.

chinmayshah99:
> @chinmayshah99 Apologies if I'm missing something obvious here, but it looks like a lot of tests are failing because the matplotlib import isn't working. Do I need to do something in my PR to address that?

Yeah, you need to add a line to [test.yml](https://github.com/OpenMined/PyDP/blob/dev/.github/workflows/tests.yml) on line 83. 
` pipenv install matplotlib --ignore-pipfile --skip-lock`. That should do the trick. 


> I will lint and edit `carrots_demo.ipynb` to improve it if @chinmayshah99 thinks it's a good idea. I've already spotted a few things (e.g. We have section `1.1` but nothing else is numbered, so we can remove it).

@8bitmp3 please go ahead.

chinmayshah99:
To run mypy in jupyter NB, here's how to do it: https://stackoverflow.com/questions/58976685/type-check-jupyter-notebooks-with-mypy

chinmayshah99:
@NateSolon this looks good enough to merge it. Adding mypy can be taken as a separate issue.

chinmayshah99 (approved):

NateSolon:
> @NateSolon this looks good enough to merge it. Adding mypy can be taken as a separate issue.

Thanks! I will learn about mypy this week.

NateSolon:
@8bitmp3 Thanks for the feedback! I'm not ignoring it, I just haven't been able to get around to addressing all of it yet!

8bitmp3:
@NateSolon No worries, we can follow-up in a separate PR 👍 Nice job by the way 💯

Issue #292, opened 2020-09-13
Get rid of Pipenv


alejandrosame:
## What?
For our use case, Pipenv introduces several problems for little gain. We should check how to substitute it.

## Why?
Substituting pipenv means that we need to change code in several setup/admin files. The options considered are:

- [ ] Check moving to Poetry -> #277 
- [ ] If Poetry proves to be also problematic, fallback to **pip+venv** and provide Make targets to make them easy to use and standard.

## Additional Context
This issue is best to be worked on after #289 is solved.

BrendanSchell:
@alejandrosame can I take this on?

alejandrosame:
@BrendanSchell You are more than welcome to take this task :) If you agree, as it is suggested in the description, we can assign you #277 and you can focus on updating the Makefile to substitute pipenv with poetry to begin with.

I hope to not take too long with #289 so these changes can be tested more seamlessly, but is not a hard requirement, it's mostly for convenience. Otherwise, once you make the Makefile targets work with poetry, further updates will be needed on different files around the repository.

BrendanSchell:
@alejandrosame That sounds good to me - I can take on #277 for now to update the Makefile to try out poetry.

alejandrosame:
Hi @BrendanSchell, how is the task going? 

If you have any doubt, do not hesitate to contact us also through Slack, we are quite active there.

BrendanSchell:
@alejandrosame thanks for checking in! Made the change to the makefile and should have some time to test out later today.

mhosankalp:
Hi,

Can I take this one or it is already been worked upon?

Thanks and Regards,
Sankalp

alejandrosame:
Hi @mhosankalp, BrendanSchell is already working on this task!

If you don't find a task to work on, you can always contact us on Slack to see how can we guide you.

mhosankalp:
Thanks @alejandrosame. I will take another one

chinmayshah99:
#310 solves this.

Issue #277, opened 2020-08-28
Moving from Pipenv to Poetry


chinmayshah99:
## Description
Poetry is a much better package manager compared to pipenv, with similarity to package manager in rust & npm. We could migrate. Open to discussions. 

## Additional Context
Poetry: https://python-poetry.org/

alejandrosame:
I'm ok with making a test branch with Poetry tooling to see if it works better for us. Pipenv certainly isn't up to what they promise (yet), it gets too much in the way for little gain.

I don't discard either falling back to plain venv and simply providing targets/actions with our Makefile.

chinmayshah99:
Sounds lika a plan.

Regards

Chinmay Shah

alejandrosame:
@BrendanSchell will work on this

PR #310, opened 2020-10-12
Replace pipenv with poetry in makefile

BrendanSchell:
## Description
As per #277 and #292, this is to replace pipenv with poetry for managing the virtual environment and python dependencies.

## Affected Dependencies
List any dependencies that are required for this change.

## How has this been tested?
tested with  
* `make build` 
* `poetry run python examples/carrots_demo/carrots.py`
* `poetry run pytest tests`
* passes github actions

## Checklist


alejandrosame:
@chinmay poetry states incompatibility with Python 3.5 (https://pypi.org/project/poetry/), is this a strong enough reason to drop poetry?

It seems that v1.0 is compatible with 3.4+ but not v1.1. I'm still trying to see what is the reason to drop support for Python 3.5 from within the same major version, but it might take me some time.

BrendanSchell:
@alejandrosame black is also incompatible with python <3.6 which is why some of these tests are failing right now. Can try figuring a workaround out, but maybe I'll just wait on @chinmayshah99 to see if it makes sense to even worry about it if poetry and python 3.5 incompatibility might be a dealbreaker anyway

alejandrosame:
@BrendanSchell Interesting, if I'm not mistaken Poetry1.0 supports Python3.5 so we could fix that version. However, I find problematic that Black is also failing for Python3.5 (so we could also pin the Black version that was working with it). I need to check what is going on with these libraries, in case there is a trend here.

chinmayshah99:
Black is a dev dependency and it's alright if it does not work with py35. 
Black is for checking style and even if it fails in any one of the other py versions, it's alright.

As far as poetry not working for py35, we need to explore that.

pyproject.toml
```diff
@@ -0,0 +1,27 @@
...
+name = "pydp"
+version = "0.1.0"
+description = ""
+authors = ["Chinmay Shah <chinmayshah3899@gmail.com>"]
+license = "Apache-2.0"
+
+[tool.poetry.dependencies]
+python = "^3.5"
+jupyter = "*"
+pandas = "*"
```
chinmayshah99: Isn't pandas, jupyter supposed to be a dev dependency?

chinmayshah99 (commented):
For the py38 OSX issue, I think this should help: https://scikit-learn.org/stable/developers/advanced_installation.html#compiler-macos

pyproject.toml
BrendanSchell: possibly, but not in the original pipfile:
```[[source]]
name = "pypi"
url = "https://pypi.org/simple"
verify_ssl = true

[dev-packages]
setuptools = "*"
bumpversion = "==0.5.3"
pytest = "*"
black = "==19.10b0"
twine = "*"
sphinx = "*"
sphinx-rtd-theme = "*"
gcovr = "*"
coverage = "*"
mypy = "*"

[packages]
jupyter = "*"
pandas = "*"

File: pyproject.toml [hunk 2]
```diff
@@ -0,0 +1,27 @@
...
+
+[tool.poetry.dev-dependencies]
+setuptools = "*"
+bumpversion = "==0.5.3"
+pytest = "*"
+black = {version = "==19.10b0",python = "^3.6"}
+twine = {version = "*", python = "^3.6"}
```
chinmayshah99: twine is required to upload the package, so we need it for py35 also

BrendanSchell: sounds good, I'll remove that python version specification for it

BrendanSchell:
@chinmayshah99 got this to work for all tests by just using a different pandas version for python 3.8 that doesn't have to compile.

chinmayshah99 (approved):

alejandrosame (approved):

PR #331, opened 2020-11-16
Subsample and aggregate Example 

kamathhrishi:
Example of subsample and aggregate DP mechanism on Carrots example.

### DIFF
--- a/examples/carrots_demo/carrots_subsampleaggregate.ipynb
+++ b/examples/carrots_demo/carrots_subsampleaggregate.ipynb
@@ -1,0 +1,639 @@
+{
+ "cells": [
+  {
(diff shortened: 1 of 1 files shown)


chinmayshah99:
1. Can you explain in the notebook a bit of what exactly the technique is?
2. Also, can you take a look at this: https://github.com/OpenMined/PyDP/blob/dev/examples/Tutorial%204-Launch_demo/DP%20proof.ipynb
3. We have something called merge(): https://pydp.readthedocs.io/en/latest/pydp.html#pydp.algorithms.laplacian.BoundedMean.merge. I'm wondering can we do a comparative analysis on the same. if there is a significant difference, we can take it up ahead

Issue #291, opened 2020-09-13
Move pydp folder under src


alejandrosame:
## What?
Python code now lives under **./pydp**, leaving our source code distributed over 2 folders.

## Why?
All source code should be under **./src** folder. In order to have a clean repository, we might need to restructure folders inside **./src** and refactor the way the bindings are made. However, it shouldn’t be too much of a drastic change (mostly folder renaming, with its corresponding bazel configuration changes).

chinmayshah99:
Note: This would require playing with bazel configuration. Familiarity with bazel (or willingness to learn about bazel features) is a must.

siddhesh1598:
Can I work on this? I am not familiar with bazel, but I would like to learn and work on it.

chinmayshah99:
Sure, hit us on slack channel `#lib_pydp` if you need anything

siddhesh1598:
Alright, will do.

chinmayshah99:
hi @siddhesh1598, how is it coming along?

siddhesh1598:
Hey @chinmayshah99, so far so good. Started learning the bazel config and cloned the repo. Will soon make changes to it.

alejandrosame:
Fixed on #332.

PR #332, opened 2020-11-19
Moved pydp folder under src

siddhesh1598:
## Description
Moved the pydp folder under the src folder. Created a BUILD for the pydp folder and extended it with the bindings_test. Closes issue #291.

## How has this been tested?
I tested it by running the docker image. I ran the following commands to reproduce the results.
```$ docker build -t pydp:test .```
```$ docker run --rm -it pydp:test```
```$ make test```
```$ make build```
```$ poetry run python examples/carrots_demo/carrots.py```

## Checklist
- [Yes ] I have followed the [Contribution Guidelines](https://github.com/OpenMined/.github/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/OpenMined/.github/blob/master/CODE_OF_CONDUCT.md)
- [Yes ] I have commented my code following the [OpenMined Styleguide](https://github.com/OpenMined/.github/blob/master/STYLEGUIDE.md)
- [Yes ] I have labeled this PR with the relevant [Type labels](https://github.com/OpenMined/.github/labels?q=Type%3A)
- [Yes ] My changes are covered by tests

### CODE REVIEW

**Review by chinmayshah99 — CHANGES_REQUESTED** [review]
Can you remove the poetry.lock and the bin file?

**Review by chinmayshah99 — CHANGES_REQUESTED** [review]
Can you please remove the BIN file?

**Comment by alejandrosame** [comment]
@siddhesh1598 ping

@chinmayshah99 I already have a testing improvement waiting for this PR to be merged, I noticed the pydp binary there and deleted it.

**Comment by chinmayshah99** [comment]
@alejandrosame 
Maintainers' edit isn't enabled on this PR. Let's wait till tonight, and then merge ( we can delete the bin file)

**Comment by siddhesh1598** [comment]
@chinmayshah99 are you talking about the bazel bin files? 
Also, the maintainer's edit is allowed on this PR.
Sorry, I was caught up with some tasks from work.

**Comment by chinmayshah99** [comment]
@siddhesh1598 can you please delete the bazel  bin file?

**Review by alejandrosame — COMMENTED** [review]
I shouldn't review my own changes :P From where Sid got stuck, I applied required changes to make it work in my local setup and in the CI environment. Now even if we rollback from poetry (or for what it matters, change to any other env manager) tests will fail if pydp isn't properly installed in the environment.

**Review by chinmayshah99 — DISMISSED** [review]

**Review by BenjaminDev — CHANGES_REQUESTED** [review]
I have a few concerns (at least on macOS but I suspect on all platforms). The project builds fine but when running `python setup.py bidet_wheel` the package_data is not correctly copied to the wheel. This means the resulting wheel does not contain the `_pydp.so` file and fails to import. 

To test locally build the wheel `python setup.py bidet_wheel` and then install that into a completely separate virtual env. Then check it imports and runs as expected.

To fix: add to `setup.py`
```
package_dir={'':'src'},

```
and move tests into `src/`

These are rushed suggestions and I might be mistaken but please confirm that the resulting wheel is complete - it's broken on my side macOS. The logs in the CI make me think it's broken there too.

Otherwise it looks good to me!

alejandrosame (commented):
In order to address the comment from @BenjaminDev I first changed our CI workflows to separate wheel build and installation in separate virtual machines. This makes it a more realistic test as it is closer to users downloading the wheel and installing (we could follow a similar approach to test wheels from pypi test instance).

After doing that, Poetry was giving problems with Python 3.5 and pytest incompatibilities, so I changed the workflows to run separately the test for 3.5. Pip will also drop support for python3.5 soon, but this allows us to keep supporting it via wheel installation until we properly communicate to users that it will no longer be supported (out of scope for this PR).

After the previous issue was fixed, it the error raised by @BenjaminDev could be replicated in CI. That's when I introduced the fix to properly refer to the src folder in setup.py.

The last commit of this batch consists on a fix that appears to happen also locally between macOS 10.15 and pip+wheel+setuptools, where the packaging tooling incorrectly identifies the system as macosx11 and breaks the build and installation workflow (it improperly names the resulting wheel).

BenjaminDev (approved):

chinmayshah99:
We would have just one last release with py35 in the coming week(s), post
which we won't be supporting py35.

Regards

Chinmay Shah

PR #338, opened 2021-01-09
Deprecate support for Python 3.5 and below

### LINKED ISSUE #334: Notify users about Python 3.5 deprecation.
## Where?
We should let user know that we are dropping support for Python3.5 after the next planned release. At least this warning should appear:
 - in the PyDP pipy page (https://pypi.org/project/python-dp/)
 - as an interactive warning in console when the user does `import pydp` when Python version is 3.5.

## Who?
PyDP package users.

## What?
PyDP users should be encouraged to upgrade their Python to 3.6+ before we stop publishing wheels for it after the next release is out for a smooth transition.

## Additional Context
Many packages are dropping support for Python 3.5 on January 2021, including `pip`which makes this python version out of the common distribution pipelines.
 
We can use as inspiration how other packages communicate Python deprecation like https://pypi.org/project/poetry (both in the pipy page and in the console output).

festusdrakon:
Closes #334


chinmayshah99:
@festusdrakon thanks for the PR, but this PR is essentially dropping support to py35. 

What we need to do is add a warning when you use py35 on machines. So in the init.py file, we can check for py35 and raise a deprecation warning. This could go as a new PR.


Changes in the PR needed:
Remove py35 from our ci-cd files. 

Note: We would be merging this PR when we drop py35 support.

festusdrakon:
Okay, I'll create a new branch with only the warning and the documentation change and create a new PR from that.

chinmayshah99:
Sounds good. Also please keep in mind that this is quite urgent.

PR #349, opened 2021-04-01
Jupyter Notebook for PyDP & Scikit-Learn comparison 

grilhami:
## Description
Added a Jupyter Notebook that compares the accuracy result of training a dataset using Scikit-Learn's `GaussianNB` and PyDP's `GaussianNB`; the latter which adding differential privacy mechanisms. 

## Affected Dependencies
`matplotlib` was added in the `pyproject.toml` file to allow the notebook to function properly.

## Checklist


pyproject.toml
```diff
@@ -7,6 +7,7 @@ license = "Apache-2.0"
 
 [tool.poetry.dependencies]
 python = "^3.6"
+matplotlib = "^3.3.4"
```
chinmayshah99: is this necessary over here? If not, we can move it to dev dependencies

chinmayshah99 (changes requested):

pyproject.toml
grilhami: It's to make sure that the notebook runs well. But moving it to dev dependencies also sounds good, I'll make the changes

chinmayshah99: Moving it to the main dependency means the user requires this to install matplotlib when using pydp. This is not desirable behavior.

grilhami: I've already moved matplotlib as a dev dependency

chinmayshah99 (approved):

PR #353, opened 2021-04-13
Boston Housing Price Demo added under examples folder

msulemannkhan:
## Description
This is an example for PyDP using Boston housing price dataset

## Affected Dependencies
No

## How has this been tested?
Manually verified

## Checklist


chinmayshah99 (changes requested):
Can we not add dataset from sklearn and directly call it from source?
or maybe call from a URL?
Calling it from sklearn means sklearn needs to be present

msulemannkhan:
Right, I have added the data file.

Frank-7 (changes requested):
If we are going to read the dataset locally there is no need for this line on the second cell:
from sklearn.datasets import load_boston

Please remove that line and I will approve the change

msulemannkhan:
Right, I have updated removed this line.

Frank-7 (approved):
Thanks, that would do it

PR #359, opened 2021-05-02
added documentation for PyDP naive bayes comparison

dnabanita7:
## Description
added docs for PyDP comparison

## Checklist


grilhami:
@dnabanita7 Great work on adding those descriptions! However, I would like to suggest some minor adjustments.

1.  I think it would be a good idea to mention the description of some variables in a code block like `this`. For example, instead of just "clf", we can do `clf`. I think that would help make the description more readable.

2.  For the part where `mean_train_accs`,  `mean_test_accs`, and `dp_results`, it'll be great if we can specify that those results are differentially private since the `GaussianNB` class from PyDP is differentially private.

dnabanita7:
cc @chinmayshah99 @grilhami

chinmayshah99 (approved):

Issue #358, opened 2021-04-28
1.0.2 Wheel wont install on MacOS 10.15


madhavajay:
## Description
I just noticed that the MacOS Wheels for `1.0.2` won't install on my MacOS `10.15`.

This means the build process isn't quite right because wheels should be backwards compatible with MacOS 11 and 10.15 and even older.

Renaming the file to: `python_dp-1.0.2-cp38-cp38-macosx_10_15_x86_64.whl` works which is a known solution for these incorrectly built wheels.

I believe @bcebere fixed this on TenSEAL back in December.

## Are you interested in working on this improvement yourself?
Possibly

## Additional Context
None needed.

bcebere:
I think the issue is for Bazel environments only.

Here is the fix for the PSI project, if I remember correctly https://github.com/OpenMined/PSI/blob/master/.github/workflows/scripts/release_osx.sh

You need to set the MACOSX_DEPLOYMENT_TARGET env and build the wheel using
```
python3 setup.py build bdist_wheel --plat-name macosx_10_14_x86_64 --dist-dir wheel
```

victorperezc:
Hi @bcebere! I was facing the same installation issue as @madhavajay on macOS Big Sur Version 11.2.3. Fixed it by following your instructions above. Nevertheless, when I import `pydp` now I got the following error

```
<ipython-input-14-401ae74f6323> in <module>
      9 logging.basicConfig(level=logging.ERROR)
     10 
---> 11 import pydp
     12 sy.load('pydp')

/usr/local/lib/python3.9/site-packages/pydp/__init__.py in <module>
----> 1 from pydp import algorithms, distributions, util
      2 
      3 
      4 __version__ = "1.0.2"
      5 

/usr/local/lib/python3.9/site-packages/pydp/algorithms/__init__.py in <module>
----> 1 from . import laplacian
      2 
      3 __all__ = ["laplacian"]

/usr/local/lib/python3.9/site-packages/pydp/algorithms/laplacian/__init__.py in <module>
----> 1 from ._bounded_algorithms import (
      2     BoundedMean,
      3     BoundedSum,
      4     BoundedStandardDeviation,
      5     BoundedVariance,

/usr/local/lib/python3.9/site-packages/pydp/algorithms/laplacian/_bounded_algorithms.py in <module>
----> 1 from .._algorithm import BoundedAlgorithm
      2 
      3 
      4 class BoundedMean(BoundedAlgorithm):
      5     """

[...]

PR #363, opened 2021-05-11
Fixing CI

madhavajay:
## Description
Fixing CI.

## Affected Dependencies
None

## How has this been tested?
Locally and with CI.

## Checklist


chinmayshah99:
Does this also solve #358 ?

madhavajay:
> Does this also solve #358 ?

Kind of. So it will build and install in CI, but really we should make sure that the wheel it creates can also install on any version of MacOS. There might be some extra tricks to making that happen, and we will likely need to add a second entry to the matrix to be sure:

```
macOS Big Sur 11.0 | macos-11.0
macOS Catalina 10.15 | macos-latest or macos-10.15
```

madhavajay:
I also think this should be done a different way:
https://github.com/OpenMined/PyDP/blob/dev/tests/notebooks/test_all_notebooks.py

But before refactoring it would be good to get a clear understanding of the goal? I would highly suggest running linting and checks before committing in pre-commit and then again in the linting step of CI, but currently the pre-commit checks aren't all passing because there are a lot of flake8 issues.

Try this:
```
$ pre-commit run --all-files
Found 216 errors in 29 files (checked 42 source files)
```

However, I think they should be handled as a seperate set of tasks so for now the key is just to get the tests passing and the build working so that we can do a Python 3.9 release.

Once these are passing for 3.9 I can add a nightly build to run the older versions like 3.6, 3.7 and 3.8 every night, and a simple file which can be used to trigger them manually by doing something like this:
```
$ date > tests/trigger/versions
```

.pre-commit-config.yaml
```diff
@@ -0,0 +1,57 @@
...
+      - id: trailing-whitespace
+      - id: check-docstring-first
+      - id: check-json
+      - id: check-added-large-files
+      - id: check-yaml
+      - id: check-merge-conflict
+      - id: check-executables-have-shebangs
+      - id: debug-statements
+      - id: name-tests-test
+        exclude: ^(tests/syft/lib/allowlist_report.py)
```
madhavajay: remove

replomancer:
I only investigated `test_bounded_mean.py` There may be some issues with other tests (they fail for me). Here's what I found so far.

This test:
https://github.com/OpenMined/PyDP/blob/c387752d2b42aebdce52f8fe687d0d9aab828b45/tests/algorithms/test_bounded_mean.py#L69-L73
means if you compute a dp mean of many many many fives, the result should be close to five.

Apparently for performance reasons, instead of building big lists of fives, an attempt is made here to load data:
https://github.com/OpenMined/PyDP/blob/c387752d2b42aebdce52f8fe687d0d9aab828b45/tests/algorithms/test_bounded_mean.py#L49-L62

from here:
https://github.com/OpenMined/PyDP/blob/c387752d2b42aebdce52f8fe687d0d9aab828b45/tests/algorithms/test_bounded_mean/test_bounded_mean_int64_data.proto#L1-L2
This file looks a little different when something useful is serialized here. Currently instead of having many many fives, BoundedMean probably gets nothing and produces random results between lower and upper bounds. This test passes after data file is deleted but it takes almost 2 minutes on my machine.

~Speaking of lower and upper bounds. Maybe something builds differently for me or somehow I downloaded another C++ lib version but for me this is also broken:~ (this was a result of some dev changes I was missing)
https://github.com/OpenMined/PyDP/blob/c387752d2b42aebdce52f8fe687d0d9aab828b45/tests/algorithms/test_bounded_mean.py#L47
~i.e. I get~
```
E       RuntimeError: LInf sensitivity has to be positive but is 0
```
~because~
```python
        x = BoundedMean(1.0, 0, 0, 10, dtype="int64")
```
~means~
```python
        # lower_bound same as upper_bound:
        x = BoundedMean(epsilon=1.0, lower_bound=0, upper_bound=0, l0_sensitivity=10, dtype="int64")
```
~I also get the same sensitivity error from here:~
https://github.com/OpenMined/PyDP/blob/c387752d2b42aebdce52f8fe687d0d9aab828b45/tests/algorithms/test_bounded_mean.py#L77-L78
~I need to increase upper_bound for it to stop crashing.~

I may have ideas/comments tomorrow.

replomancer:
(edit: this was from missing some dev changes)
~Other fails I see also result from sensitivity set to zero (LInf sensitivity is computed from other parameters) e.g.~
https://github.com/OpenMined/PyDP/blob/c387752d2b42aebdce52f8fe687d0d9aab828b45/tests/algorithms/test_bounded_variance.py#L6-L11
~Here~
```python
BoundedVariance(epsilon, delta, lower_bound, upper_bound, dtype="float") 
```
~actually means~
```python
BoundedVariance(epsilon=epsilon, lower_bound=delta, upper_bound=lower_bound, l0_sensitivity=upper_bound, dtype="float") 
```
~so both `lower_bound` and `upper_bound` are set to `0`.~

~Similar story in `TestBoundedStandardDeviation` and `TestBoundedSum`.~

.bazelrc
```diff
@@ -1,3 +1,3 @@
-build:Linux --cxxopt='-std=c++17'
-build:macOS --cxxopt='-std=c++17'
-build:Windows --cxxopt='/std:c++17'
+build:linux --cxxopt='-std=c++17'
+build:macos --cxxopt='-std=c++17' --features=-supports_dynamic_linker
```
chinmayshah99, line 2: Is there a reason why these flags are added to macOS only?

.github/workflows/pypipublish_linux.yml
```diff
@@ -54,7 +54,7 @@ jobs:
 
       - name: Build pydp lib
         run: |
-          bazel build --config Linux src/python:bindings_test  --verbose_failures
+          bazel build --config ${{ matrix.os }} src/python:pydp  --verbose_failures
```
chinmayshah99: I am not sure if `matrix.os` would pick OS since this is hardcoded. Is this verified?

build_PyDP.sh
```diff
@@ -4,11 +4,11 @@
 source ./get_platform.sh
 
 case $PLATFORM in
-  *"linux"*)
-    PLATFORM="Linux"
+    *"linux"*)
+        PLATFORM="Linux"
```
chinmayshah99: can we change this to: `PLATFORM="linux"`.
Needed due to change over here: https://github.com/OpenMined/PyDP/blob/c387752d2b42aebdce52f8fe687d0d9aab828b45/.bazelrc

.bazelrc
madhavajay: Yes, because they are only needed for macOS

.github/workflows/pypipublish_linux.yml
madhavajay: I have actually refactored these now that the tests pass anyway, so its going to be identical to the existing tests yaml which works.

tests/algorithms/test_bounded_mean.py
```diff
@@ -62,9 +68,12 @@ def _make_loaded_object(value, size, iter):
     return _make_loaded_object
 
 
+# uses: ./tests/algorithms/test_bounded_mean/test_bounded_mean_int64_data.proto
 def test_bounded_mean_int64(make_loaded_object):
     x = make_loaded_object(5, 100000000, 5)
-    assert expect_near(5.0, x.result(), 0.1)
+    res = x.result()
+    print(x, type(x), res, type(res))
```
chinmayshah99: remove this print statement from here?

tests/algorithms/test_bounded_mean/test_bounded_mean_int64_data.proto
```diff
@@ -1,2 +1,2 @@
-
-;type.googleapis.com/differential_privacy.BoundedMeanSummary
+M
+;type.googleapis.com/differential_privacy.BoundedMeanSummary�ʵ���
```
chinmayshah99: some extra characters are added to the file. Verify if this is really needed?

prereqs_linux.sh
```diff
@@ -22,7 +22,7 @@ else
     sudo apt-get install python3.6
 fi
```
chinmayshah99, line 28: change the SHA on line 62

chinmayshah99 (changes requested):

prereqs_linux.sh
chinmayshah99: Also, this needs an update in prereqs_osx.sh

tests/algorithms/test_bounded_mean/test_bounded_mean_int64_data.proto
replomancer: This is needed for `test_bounded_mean_int64`.

This file has a very confusing name. It's called `test_bounded_mean_int64_data.proto`, but it doesn't store a protobuf message type definition, instead it actually stores some serialized data. The reason Madhava's PR was not passing checks was because this thing was missing here. If you restore the old version of this file (remove weird characters), you will see `test_bounded_mean_int64` failing again.

chinmayshah99 (changes requested):
LGTM! can you remove the tunnel ssh?

madhavajay:
Versions less than 3.9 are now building correctly and all tests are passing: https://github.com/OpenMined/PyDP/actions/runs/875497191

chinmayshah99 (approved):

PR #364, opened 2021-05-16
[WIP] Quickfix CI

replomancer:
Just to save time here's a small quickfix for #363 in a separate PR.

I rebased the whole thing on current dev.

Issues were:
- ~lower and upper bounds can't be too similar or else sensitivity becomes zero and we get errors (see comments in #363 )~ (this was actually a result of mismatches between branches, hopefully solved by rebasing,
- serialized data was missing (the file existed, but data wasn't there)

I tested this locally, let's see if checks pass here.


chinmayshah99:
If you refer to changes made in #361, when we upgraded the SHA, we added delta as a compulsory parameter.

PR #374, opened 2021-07-02
Adding `dp::PartitionSelectionStrategy` wrapper for Python

levzlotnik:
## Description
As per [the issue on PipelineDP](https://github.com/OpenMined/PipelineDP/issues/52), adding support for private partition selection strategies:
1. Truncated geometric thresholding ([paper](https://arxiv.org/pdf/2006.03684))
2. Laplacian/Gaussian thresholding ([ref](https://github.com/google/differential-privacy/blob/main/common_docs/Delta_For_Thresholding.pdf))

The API is described in [the issue on PipelineDP](https://github.com/OpenMined/PipelineDP/issues/52).

## Affected Dependencies
No affected dependencies. Some bazel BUILD files have been changed to include `@google_dp//...:partition-selection`.

## How has this been tested?
The following python script reproduces [Figure 1 from the Paper](https://arxiv.org/pdf/2006.03684):
```python
import numpy as np
import matplotlib.pyplot as plt

import pydp as dp

epsilon = 0.1
delta = 1e-10
num_partitions = 1
amount_of_users = np.arange(150, 301)
# amount_of_users = np.arange(5, 21)
truncated_geometric = dp.partition_selection.create_partition_strategy(
    "truncated_geometric", epsilon, delta, num_partitions)
laplace = dp.partition_selection.create_laplace_partition_strategy(epsilon, delta, num_partitions)

probs_truncated_geometric = []
probs_laplace = []
for num_users in amount_of_users:
    sims_truncated_geometric = []
    sims_laplace = []
    for sim in range(10000):
        sims_truncated_geometric.append(truncated_geometric.should_keep(num_users))
        sims_laplace.append(laplace.should_keep(num_users))
    probs_truncated_geometric.append(np.mean(sims_truncated_geometric))
    probs_laplace.append(np.mean(sims_laplace))

plt.plot(amount_of_users, probs_truncated_geometric, color='red')
plt.plot(amount_of_users, probs_laplace, color='blue', linestyle='dashed')
plt.xlabel("$n$")
plt.ylabel(r"$\mathbb{P}[\rho(n) = keep]$")
plt.xlim(amount_of_users.min(), amount_of_users.max())
plt.ylim(0, 1)
title = "Release Probability depending on \nthe number of unique users. $\\varepsilon=%.1e, \\delta=%.1e$" % (
    epsilon, delta)
plt.title(title)
plt.show()

```

## Checklist


src/bindings/PyDP/algorithms/partition_selection_strategies.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include "pybind11/stl.h"
+
+#include "algorithms/partition-selection.h"
+
+using namespace std;
+
+namespace py = pybind11;
+namespace dp = differential_privacy;
+
+typedef std::unique_ptr<dp::PartitionSelectionStrategy> PyPartitionSelectionStrategy;
```
dvadym: Nit: using PyPartitionsSelectionStrategy =  ...
https://google.github.io/styleguide/cppguide.html#Aliases

src/bindings/PyDP/algorithms/partition_selection_strategies.cpp
```diff
@@ -0,0 +1,72 @@
...
+  dp::PreaggPartitionSelection::Builder builder;
+  builder.SetEpsilon(epsilon);
+  builder.SetDelta(delta);
+  builder.SetMaxPartitionsContributed(max_partitions_contributed);
+
+  auto obj = builder.Build();
+  if (!obj.ok()) {
+    throw std::runtime_error(obj.status().ToString());
+  }
+  return std::move(obj.ValueOrDie());
```
dvadym, line 26: Nit (style improvement): no need in `std::move`, the value in `return` is automatically r-value, and automatically will be moved.

This same in other places

src/bindings/PyDP/algorithms/partition_selection_strategies.cpp
```diff
@@ -0,0 +1,72 @@
+// Provides bindings for Bounded Functions
+
+#include "pybind11/complex.h"
+#include "pybind11/functional.h"
+#include "pybind11/pybind11.h"
+#include "pybind11/stl.h"
+
+#include "algorithms/partition-selection.h"
+
+using namespace std;
```
dvadym: Nit  (style improvement): remove using namespace std;

https://google.github.io/styleguide/cppguide.html#Namespaces

src/bindings/PyDP/algorithms/partition_selection_strategies.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+namespace py = pybind11;
+namespace dp = differential_privacy;
+
+typedef std::unique_ptr<dp::PartitionSelectionStrategy> PyPartitionSelectionStrategy;
+
+PyPartitionSelectionStrategy create_truncted_geometric_partition_strategy(
```
dvadym: Nit: CreateTruncatedGeometric....
https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

(as far as I understand C++ names and Python names can be different)

src/bindings/PyDP/algorithms/partition_selection_strategies.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+PyPartitionSelectionStrategy create_laplace_partition_strategy(
+    double epsilon, double delta, int max_partitions_contributed
+    ) {
+  dp::LaplacePartitionSelection::Builder builder;
+  builder.SetEpsilon(epsilon);
+  builder.SetDelta(delta);
+  builder.SetMaxPartitionsContributed(max_partitions_contributed);
+  // builder.SetLaplaceMechanism(std::move(laplace_builder));
```
dvadym: Nit: drop the commented code

src/pydp/algorithms/partition_selection.py
```diff
@@ -0,0 +1,16 @@
+from .._pydp import _algorithms
+
+PartitionSelectionStrategy = _algorithms.PartitionSelectionStrategy
+PartitionSelectionStrategy.__doc__ = """
+Provides a common abstraction for PartitionSelectionStrategy. Each partition
+selection strategy class has a builder with which it can be instantiated, and
```
dvadym: In this PR in Python interface there are factory functions instead of builders, so maybe something line
" has a factory function. " 
And to list somewhere the list of all factory functions

dvadym (commented):
Thanks it looks great. Just some style comments

src/bindings/PyDP/algorithms/partition_selection_strategies.cpp
levzlotnik: Apparently - `obj` is of type `base::StatusOr<std::unique_ptr<...>>`, and the method `obj.ValueOrDie()` is overloaded. 
The overloads are:  
```
  const T& ValueOrDie() const&;
  T& ValueOrDie() &;
  const T&& ValueOrDie() const&&;
  T&& ValueOrDie() &&;
```
From [here](https://github.com/google/differential-privacy/blob/114505ef7d70cc25cd5472461d80e51f5ed1c4a8/cc/base/statusor.h#L207-L210). 
The return type is a value type, so the compiler chooses the first version of `const T& ValueOrDie() const&;`, which fails the compilation because `std::unique_ptr` cannot get copied from const ref.

src/pydp/algorithms/partition_selection.py
levzlotnik: Used `.pyi` file instead for interface.

src/pydp/partition_selection/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+
+    Parameters
+    --------------
+    strategy: str
+        One of:
+            * **'truncated_geomteric'**: creates a `Truncated Geometric <https://arxiv.org/pdf/2006.03684.pdf>`_ Partition Strategy.
+            * **'laplace'**: creates a private partition strategy with Laplace mechanism.
+            * **'gaussian'**: creates a private partition strategy with Gaussian mechanism.
+
+    epsilon: float
```
dvadym: Nit: no need to indicate types here, since type annotations are used

https://google.github.io/styleguide/pyguide.html#doc-function-args

src/pydp/partition_selection/__init__.pyi
```diff
@@ -0,0 +1,48 @@
+class PartitionSelectionStrategy:
```
dvadym: Nit: you can use [abc](https://docs.python.org/3.6/library/abc.html) for making the class abstract (though I'm not sure how abc interplays with Pybind11)

src/bindings/PyDP/algorithms/partition_selection_strategies.cpp
dvadym: Wow, thanks, good to know :)

src/bindings/PyDP/algorithms/partition_selection_strategies.cpp
```diff
@@ -0,0 +1,79 @@
...
+  auto obj = builder.Build();
+  if (!obj.ok()) {
+    throw std::runtime_error(obj.status().ToString());
+  }
+  return std::move(obj.ValueOrDie());
+}
+
+PyPartitionSelectionStrategy CreateLaplacePartitionStrategy(
+    double epsilon, double delta, int max_partitions_contributed
+    ) {
```
dvadym: Nit: is it right formating?

dvadym (dismissed):
Thanks for implementing!

Approve from correctness and usage of Google C++ DP library point of view.

src/pydp/partition_selection/__init__.pyi
levzlotnik: This file is for intellisense only (like Jedi or Pylance for vscode). 
It doesn't interact with real python execution.

chinmayshah99: Can we rename this file to `.py` file? We have not used pyi files till now

src/pydp/__init__.py
```diff
@@ -6,6 +6,7 @@
 from pydp import distributions
 from pydp import util
 from pydp import ml
+from pydp import partition_selection
```
chinmayshah99: Rather than creating it as the main module, we could move it inside algorithm as a submodule. 
Thoughts @dvadym?

chinmayshah99 (changes requested):
Need to add tests for this new addition

chinmayshah99:
Also, style tests are failing. Please look into that. You can run `make` to format both python and cpp code

src/pydp/partition_selection/__init__.pyi
levzlotnik: Moved interfaces into `.py` file. This doesn't interfere with pybind11.

src/pydp/__init__.py
levzlotnik: Moved to `algorithms`.

levzlotnik:
@dvadym @chinmayshah99 
I added tests for Truncated Geometric partition selection. For Laplace/Gaussian mechanisms partition selections python tests I would require the Laplace/Gaussian mechanisms from #372.

chinmayshah99 (approved):

levzlotnik:
@chinmayshah99 @dvadym thank you for approving the PR! 😄

PR #383, opened 2021-08-10
Update dp lib v1.1.0

dibakch:
## Description

Update Google's DP Lib to version 1.1.0.  The `xor_strings` function got removed in this release.  Newly added functionality is not exposed as of now (for instance quantile trees).

This PR is based on top of a previous PR, that should be discussed and merged first.

## Affected Dependencies

Only updated the DP Library to version 1.1.0.

## How has this been tested?

  * Ran `bazel test ...` locally.


chinmayshah99:
Included this in #370

PR #384, opened 2021-08-11
Revert "Fix Bazel build"

chinmayshah99:
Reverts OpenMined/PyDP#382
This build breaks the local build process when we build using make. 
cc: @dibakch


dibakch:
Yikes.  This PR also seems to have the exact same error.

chinmayshah99:
> Yikes. This PR also seems to have the exact same error.

The windows 3.9 error right?

dibakch:
This is well, but also the error when running `make run-tests-only` in the Docker container.

PR #394, opened 2021-09-16
Fixed the absolute reference issue in bazel build

chinmayshah99:
## Description
Earlier the WORKSPACE file in google DP/cc folder started making an absolute reference to the package name which caused our systems to fail as we were using relative naming. 
This is fixed by giving the same name to the CC folder as the actual name in google DP library thus resolving this name reference issue and now bazel can solve it internally. 

## Checklist


dibakch:
Looks as if we have a setup problem for Python 3.9 on Windows, but looks good otherwise.  Thanks! :)

dibakch (approved):

PR #400, opened 2021-11-08
Removing unused code

dvadym:
This PR contains:
1. Removing 3 unused files.
2. Removing some unused import/include
3. Using `pytest.approx` instead the only used function `expect_near` from **conftest.py** (and it's better to use standard function instead of custom one)


chinmayshah99:
C++ style check is failing. You could fix by running `make`.

chinmayshah99:
C++ style check is failing. You could fix by running `make`.

dvadym:
> C++ style check is failing. You could fix by running `make`.

Thanks. I've run formatting

chinmayshah99 (approved):

PR #405, opened 2021-12-15
Update Google C++ DP library to the latest commit and fix compilation errors

dvadym:
The API change in C++ library - `privacy_budget` and `remaining_privacy_budget` are not part of Algorithm API anymore

### DIFF
--- a/src/pydp/algorithms/_algorithm.py
+++ b/src/pydp/algorithms/_algorithm.py
@@ -72,12 +72,6 @@ class MetaAlgorithm:
         """
         return self._linf_sensitivity
 
-    def privacy_budget_left(self) -> float:
-        """
-        Returns the remaining privacy budget.
-        """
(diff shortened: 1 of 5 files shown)


src/pydp/algorithms/_algorithm.py
```diff
@@ -112,29 +106,17 @@ def quick_result(self, data: List[Union[int, float]]) -> Union[int, float]:
 
     def result(
         self,
-        privacy_budget: Union[float, None] = None,
```
chinmayshah99, line 115: Can we not use a partial privacy budget anymore?

dvadym: We can't use partial privacy budget anymore. The idea is that the budget should be tracked on a higher level of DP system, not on the level of building block library.

chinmayshah99 (approved):

PR #426, opened 2022-08-05
Extending partition selection strategies wrappers

dvadym:
It contains the following new properties/methods:
  1. Exposing `epsilon`, `delta`, `max_partition_contributed` properties
  2. `probability_of_keep` method, for exact probability of keeping partition
For laplace/gaussian strategies:
  3.  `noised_value_if_should_keep` which returns (count+noise) if `count+noise >= threshold`
  4. `threshold` which is used for selecting partitions


src/bindings/PyDP/algorithms/partition_selection_strategies.cpp
```diff
@@ -27,25 +24,74 @@ PyPartitionSelectionStrategy CreatePartitionStrategy(double epsilon, double delt
...
+  // Truncated Geometric Partition selection strategy.
+  init_partition_selection_strategy<dp::PreaggPartitionSelection>(
+      m, "TruncatedGeometricPartitionSelectionStrategy",
+      "Truncated Geometric (epsilon, delta)-differenially private partition "
+      "selection strategy.");
   m.def("create_truncated_geometric_partition_strategy",
         &CreatePartitionStrategy<dp::PreaggPartitionSelection>);
+
+  // Laplace Partition selection strategy.
+  auto py_laplace_strategy_class =
```
chinmayshah99: do we need auto here? Can we directly use the `Strategy` class?

dvadym: Yeah, it makes sense to use explicit type here. Done. PTAL

chinmayshah99 (approved):

PR #427, opened 2022-08-05
Update submodule to the latest version of Google DP library and fix compilation

dvadym:
Compilation fixes: Google DP library started using `absl::StatusOr` instead of custom `StatusOr`.

**Note**: src/bindings/PyDP/base/status.cpp is deleted, because only  `absl::StatusOr` is used


src/bindings/PyDP/pydp_lib/algorithm_builder.hpp
```diff
@@ -222,19 +223,24 @@ class AlgorithmBuilder {
...
+        return dp::GetValue<int64_t>(result.value());
     });
 
     // Other methods
     pyself.def("reset", &Algorithm::Reset);
 
     pyself.def("serialize", &Algorithm::Serialize);
 
-    pyself.def("merge", &Algorithm::Merge);
+    pyself.def("merge", [](Algorithm& pythis, const dp::Summary& summary) {
```
dvadym, line 238: returning status is not Python style, it's better to raise an exception (as it's already done in other places)

chinmayshah99: is `status.toString()` user friendly?

dvadym: Yes (at least it's as friendly as for C++ library)
Example:  **INTERNAL: Bounded mean summary must have exactly one pos_sum but got 32**

chinmayshah99 (approved):

PR #433, opened 2022-10-21
Implement QuantileTree.Merge method

dvadym:
`QuantileTree.Merge` should return `Summary` as all other PyDP objects, but in the 1st implementation it returned `dp::BoundedQuantilesSummary`. This PR fixes that.

Along a way, some minor stuff fixed as well.
1.`max_partitions_contributed_to`->`max_partitions_contributed`
2. Fixed a text of one exception.


chinmayshah99 (commented):
LGTM!
Just wanted to confirm if `max_partitions_contributed_to` is not used anywhere in the example?

dvadym:
I've checked, `max_partitions_contributed_to` it's not used in examples. I've replaced it in all places in QuantileTree code.

chinmayshah99 (approved):

PR #442, opened 2023-02-04
Update to the latest version of Google DP library

dvadym:
This PR updates Google C++ library to the [latest commit](https://github.com/google/differential-privacy/commit/2a1e55484aeb9ca444a3b34dcf128d454e80f255)

### DIFF
--- a/tests/algorithms/test_partition_selection.py
+++ b/tests/algorithms/test_partition_selection.py
@@ -89,7 +89,7 @@ class TestPartitionSelection:
         ],
         [
             (10, 1, 1e-5, 1, 0.017845473615190732),
-            (1100, 0.2, 1e-7, 10, 0.007884076914531857),
+            (1100, 0.2, 1e-7, 10, 1.0),
         ],
     )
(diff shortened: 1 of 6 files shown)


tests/algorithms/test_partition_selection.py
```diff
@@ -89,7 +89,7 @@ def test_laplace_keep_and_return_noised_value(
         ],
         [
             (10, 1, 1e-5, 1, 0.017845473615190732),
-            (1100, 0.2, 1e-7, 10, 0.007884076914531857),
+            (1100, 0.2, 1e-7, 10, 1.0),
```
chinmayshah99, line 92: is there a reason to change this number?

chinmayshah99 (commented):
LGTM

tests/algorithms/test_partition_selection.py
dvadym: Yes, I made a fix in the partition selection library. It's utility fix, not privacy, from privacy stand point everything was correct, but the utility suffered. So it's expected.

PR #455, opened 2023-05-22
update README about BoundedMean function

tak-ka3:
## Description
What is written in README.md file is old. So I updated it.
First of all, the execution of `BoundedMean(0.6, 1, 10)` has an error, because the arguments of BoundedMean function are `epsilon: float = 1.0, delta: float = 0, lower_bound: Union[int, float, None] = None, upper_bound: Union[int, float, None] = None, l0_sensitivity: int = 1, linf_sensitivity: int = 1, dtype: str = "int"` in order. So the second argument of `BoundedMean(0.6, 1, 10)` is incompatible to `delta`, which should be float type not integer type. 

Also the type of epsilon should be float not double, and modify some names of arguments like from `lower` to `lower_bound`. 

## Affected Dependencies
None.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,15 +63,14 @@ import pydp as dp
 from pydp.algorithms.laplacian import BoundedMean
 
 # Calculate the Bounded Mean
-# Structure: `BoundedMean(epsilon: double, lower: int, upper: int)`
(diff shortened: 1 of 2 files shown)


chinmayshah99 (approved):

chinmayshah99:
good catch! thank you @tak-ka3 for the PR :)

PR #473, opened 2023-10-10
Build wheel with Manylinux

dvadym:
This PR implements wheel build with Manylinux for Linux.

## Why is it needed?

When a Linux binary is built , it is linked against the standard libraries (libc, libstdc++ etc) of versions that are present on the built machine. Those libraries are forward compatible, namely it's possible to run the binary on machines which have standard libraries newer than ones against which this binary was built. But it's not **backward** compatible, i.e. the binary might fail on the machines with older libraries with errors like `[usr/lib/libstdc++.so.6: version GLIBCXX_3.4.29 not found]`.

Currently since PyDP is built on Ubuntu 22, it fails to run on many older machines. We can choose Ubuntu 20, but there will be still problems with older machines.

## What is the solution?
The common solution for Python packages which have C++ parts are built C++ with [Manylinux](https://github.com/pypa/manylinux) docker images. Those are Docker images which contain old enough standard libraries, which improves compatibility.

For building [cibuildwheel](https://github.com/pypa/cibuildwheel) is used. This libraries manages build in Docker with Manylinux images.

**Note:** many common Python libs (e.g. Numpy, Matplolib) use cibuildwheel. [Here](https://github.com/pypa/cibuildwheel#working-examples)) are some other examples.

## What in this PR
It contains extending publish workflow with build on manylinux for Linux platform. Other platforms are not changed.
Namely it contains

1. Introducing `build_PyDP_linux.sh` which installs bazel and  run `build_PyDP.sh`.
2. Extending `setup.py` to call `build_PyDP_linux.sh` for Linux.
3. Updating `publish.yml` to use cibuildwheel on Linux

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -62,8 +62,20 @@ jobs:
             ./bazel-cache
           key: ${{ runner.os }}-bazel-${{ hashFiles('WORKSPACE') }}
 
-      - name: Build Google DP Unix
-        if: runner.os != 'Windows'
+      - name: Build Google DP and wheel on Linux
+        if: runner.os == 'Linux'
+        timeout-minutes: 20
+        run: |
+          pip install cibuildwheel
+          # Set for which Python version to build. 
+          py_version=${matrix.python-version//.}          
+          EXPORT CIBW_BUILD=cp${py_version}-manylinux_x86_64
+          # Build wheel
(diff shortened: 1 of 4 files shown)


chinmayshah99 (approved):
LGTM! But for linux, we're creating two bazel cache which we can avoid.

PR #477, opened 2023-10-18
Disable tests on publish and bump version

dvadym:
The current tests in publish should test the newly built wheel, but actually they run from `src/` folder. 

Tests now fail because:
Before: on the running tests, `src/pydp/_pydp.so` is used. 
Now: In the new build with `cibuildwheel`, the build happens in Docker, so `src/pydp/_pydp.so` is not present 

Anyway the tests don't test wheel, which they should test, let's for now disable them, to release a new RC candidate version and then let's fix later.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -164,6 +164,8 @@ jobs:
           python -c "import pydp; print(pydp.__version__)"
 
       - name: Run Pytest
+        if: runner.os != 'Linux'
+        # TODO(dvadym): fix tests.
         run: |
           poetry run pytest tests -n auto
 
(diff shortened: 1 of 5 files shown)


.github/workflows/publish.yml
```diff
@@ -177,11 +179,6 @@ jobs:
         run: |
           Get-ChildItem -Path ./ -Filter "*.whl" -Recurse -File | foreach {poetry run twine check $_.FullName}
 
-      - name: Renaming wheel
```
dvadym, line 180: No need to rename, it's already has a correct name with _manylinux

chinmayshah99 (approved):

PR #484, opened 2025-08-24
CMakeLists.txt build for tests and publishing

dvadym:
Google DP building block library recently introduced support of CMake. This PR implements support of Cmake build in PyDP instead of Bazel. CMake is more maintainable solution (more material, more experts).

This PR for now support Linux. Supporting of Windows and Apple Silicon will be in the subsequent CLs.

Not necessary dependencies in project.toml were removed (we should keep them minimal, because it's very error-prone - 3 OS and 5 Python versions)


src/bindings/PyDP/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
+include(FetchContent)
+
+FetchContent_Declare(
+  google_dp
+  GIT_REPOSITORY https://github.com/google/differential-privacy.git
+  GIT_TAG        main
```
dibakch: We should fix this to a commit, so that we know which one was used for creating the package

chinmayshah99 (approved):

PR #487, opened 2026-03-20
Fix mac-os build

RamSaw:
## Fix macOS Build: Resolving ABI Dynamic Lookup Collision

This pull request resolves the persistent `ImportError` / symbol not found failures that exclusively affected the macOS GitHub Actions CI runners. It ensures that the PyDP macOS python wheels reliably and natively invoke C++ bindings exactly as they do locally and on Linux/Windows target platforms.

### The Root Cause
Through our debugging, we discovered a fatal collision caused by GitHub Actions' `macos-latest` environment. The previous CI workflow instructed macOS to eagerly install Protobuf system-wide via `brew install protobuf`. 
However, **Homebrew implicitly installs Abseil globally as dynamic libraries (`.dylib`).** 

When compiling PyDP normally, the native `CMakeLists.txt` uses `FetchContent` to safely download and build `google_dp` and `abseil-cpp` statically from scratch (`.a` archives). But due to Apple's linker behavior within Pybind11's flat namespace evaluation, Apple's `ld` aggressively encountered the globally injected Homebrew dynamic Abseil signatures on the host system, stripping the expected internal symbols from our extensions under the assumption that they would organically launch at runtime. 
Naturally, because these libraries do explicitly not exist dynamically mapped into pip-installed wheels on user machines, it cleanly broke importing `PyDP`!

### The Fix
The implementation zeroes out this "Bermuda Triangle" natively and elegantly without having to compromise PyDP CMake configurations:

* **Eradicate Host System Poisoning**: Deleted `brew install protobuf` for macOS structurally from `.github/workflows/versions.yml` (and synchronized with `publish.yml` #488 separately). Because PyDP utilizes `FetchContent` to fully build the components from exact source code securely, global host installations are obsolete and actively harmful to cross-compilation!
* **Target Scoping**: Removed `--plat-name macosx_10_14_x86_64` hardcoding, leveraging native GitHub Actions arm64 `macos-latest` (M1) architecture workflows naturally.
* **Modernization & Deprecation Fixes**: Upgraded globally antiquated workflow actions (`checkout@v4`, `setup-python@v5`) to resolve Node.js 20 warnings. Replaced deprecating `::set-output` syntax precisely with `$GITHUB_OUTPUT` across matrix targets.

### Verification
The `versions.yml` pipeline reliably achieves `SUCCESS` natively across macOS runners!


dvadym (dismissed):

dvadym (approved):
Thanks!

PR #496, opened 2026-05-08
ci: Consolidate version tests and update build matrix

RamSaw:


.github/workflows/tests.yml
```diff
@@ -125,21 +104,15 @@ jobs:
         run: |
           poetry run python setup.py build bdist_wheel --plat-name macosx_10_14_x86_64
```
dvadym: Could you please drop --plat-name macosx_10_14_x86_64, that makes MacOs tests failing

dvadym (approved):
Thanks!

